PR #6, opened 2020-07-07
Added the web3-contract tutorial

albertov19:
Deploying a contract on Moonbeam and interacting with it using the web3 library.

### DIFF
--- a/getting-started/web3-contract.md
+++ b/getting-started/web3-contract.md
@@ -1,0 +1,336 @@
+---
+title: Using Web3 for Contracts
+description: Learn how to deploy Solidity-based contracts on Moonbeam with a simple script using web3.
+---
+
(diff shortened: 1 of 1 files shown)


getting-started/web3-contract.md
```diff
@@ -0,0 +1,336 @@
...
+```
+mkdir incrementer
+cd incrementer/
+npm init --yes
+```
+
+With the package.json file created, we can then install both the Web3 and the Solidity compiler packages, by executing:
+
+```
+npm install --save web3
```
crystalin, line 45: If you find it more meaningful, you can also specify the versions in the command, ex:
`npm install --save web3@1.2.9 solc@0.6.10`

albertov19: Thanks for the tip

PR #11, opened 2020-07-28
Replace git submodule init/update by short version

crystalin:


crystalin:
It does the same thing, but is easier to copy/read

PR #13, opened 2020-12-02
Include instructions for docker for mac

JSeam2:
Hi, I'm not sure about the contribution guide to the docs. I had experienced problems setting up the dockerized local node on mac. I had issues connecting to the `127.0.0.1:9944`. It turns out that docker on mac doesn't connect to localhost on the host network. As a workaround I ended up using docker-machine and used `docker-machine ssh` to port forward the docker-machine port to the actual localhost. This allows you to connect to `127.0.0.1:9944` via the Local Node option on the polkadot.js.org apps.

I have left the steps in the following documentation. Do provide any edits and proofread where necessary.

### DIFF
--- a/code-snippets/setting-up-local/dockerformac1.md
+++ b/code-snippets/setting-up-local/dockerformac1.md
@@ -1,1 +1,1 @@
-docker-machine create moonbeam
+docker run --rm --name moonbeam_standalone -p 9944:9944 -p 9933:9933 -p 30333:30333 -p 9615:9615 purestake/moonbase:tutorial-v3 /moonbase/moonbase-standalone --dev --ws-external
--- a/code-snippets/setting-up-local/dockerformac1.md
+++ b/code-snippets/setting-up-local/dockerformac1.md
@@ -1,0 +1,1 @@
+dm create moonbeam
--- a/code-snippets/setting-up-local/dockerformac1.md
+++ b/code-snippets/setting-up-local/dockerformac1.md
@@ -1,1 +1,1 @@
-dm create moonbeam
+docker-machine create moonbeam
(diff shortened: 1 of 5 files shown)


danforbes (commented):
Great contribution! I would suggest consistent proper casing for things like "Docker" and "Docker Machine". Personally I'm fine with stylized formatting like `docker-machine` but I do prefer it to be formatted as code and I also think it's best to be consistent and refer to things one way or another.

albertov19:
Once again thanks for this. I'll work on testing this on our side to double check all the procedure and merge the PR. Afterwards, I'll do some styling retouches.

purestaketdb:
> Once again thanks for this. I'll work on testing this on our side to double check all the procedure and merge the PR. Afterwards, I'll do some styling retouches.

I solved this a little differently, give it a shot. 

`docker run --rm --name moonbeam_standalone -p 9944:9944 -p 9933:9933 -p 30333:30333 -p 9615:9615 purestake/moonbase:tutorial-v3 /moonbase/moonbase-standalone --dev --ws-external`

MacOs absolutely does not let you use `--network host` so you need to be explicit about the shared ports.

JSeam2:
@purestaketdb I tried this it works too, we should stick to this instead. I'll update the documentation. The previous steps were kinda redundant now that I realize about the `--network host` issue.

albertov19:
@JSeam2 Sorry that we took so long to get back to you, I was trying a few workarounds to get a MacOS machine available so I could sort through the steps. But ultimately, @purestaketdb stepped in and found a way to avoid using Docker Machine.

JSeam2:
One more person with a Mac should try this too and verify that it works and then I think it's good to go.

albertov19:
We need to add the following flag as well to be able to connect to MetaMask

_--rpc-external_

fotescodev:
@albertov19 can we close the issue and merge this PR? I tested it locally on my mac and it works.

PR #18, opened 2021-01-20
Patch main pageModified Docs Home to add list of content on main page

albertov19:
Our docs page has grown quite significantly and I think is somewhat difficult to know where to go when you access it.

This is a first try in creating an "organized" list of content on the main page that is readable for the user. I only listed the main Ethereum tools or features that I think developers need to know.


fotescodev (approved):
Good and concise!

PR #21, opened 2021-01-25
Added the block explorer page.

fotescodev:
Added the block explorer page.


albertov19:
LGTM - Only comment would be to add key words to the "description" as this information is used for metadata (useful for searches and everything). For example, I would add:

description: An overview of the currently available block explorers that may be used to navigate the Substrate and Ethereum layers of the Moonbeam TestNet.

I'll merge and I'll add a couple of screenshots for each (also I'll add the longer description).

PR #30, opened 2021-02-10
Variable structure for new bridge to Rinkeby

mario-purestake:
Adding the variables for our new bridge to Rinkeby. They will be used later to update the Chainbridge section and include both bridges.


albertov19 (approved):
LGTM

PR #32, opened 2021-02-15
Updated to HDWalletProvider

albertov19:
Updated the Truffle tutorials so that they are now using HD-Wallet-Provider instead of our custom private-provider.

This also means the Truffle box repository was modified as well.

Everything was tested and working both in Moonbase Alpha and in a Standalone Node.


JoshOrndorff:
Awesome Work!

Just popping in to say that there is no more standalone node, and make sure you tested this against a recent moonbeam commit.

albertov19:
From a developers point of view standalone will keep existing right? There is no more the standalone binary but using the `--dev` is "more or less" a standalone node. (I understand is not entirely true but if you are a end-user running a binary with --dev flag it kind of is)

PR #63, opened 2021-04-09
Adds Blockscout and Subscan

albertov19:
This simple PR adds Blockscout and Subscan to the block explorers option.

Also revamps a little bit the entire page.


eshaben (commented):
Just a few little things but looks good! I had just opened a ticket for taking new screenshots for this page so this is awesome to see!!

PR #67, opened 2021-04-16
Update chainbridge

eshaben:
This PR includes changes to the ChainBridge documentation to reflect the new implementation. 

- Previously we had two bridges, two ERC20 handlers, two ERC721 handlers, two ERC20S, two ERC721Ms, etc one for Rinkeby and one for Kovan. The new implementation merges this so we only need one of each contract for both Rinkeby and Kovan. Now when interacting with the bridge contract, users will specify a `destinationChainID` for which network they would like to send tokens to. 
- Previously users were able to mint ERC721M tokens on Moonbase Alpha to send to Rinkeby/Kovan and then mint ERC721E tokens on Rinkeby/Kovan to send to Moonbase Alpha. Now ERC721M tokens are only mintable on Moonbase Alpha and then can be transferred back and forth to Rinkeby/Kovan.
- With all of these changes, all of the contract addresses have been updated
- There are updated screenshots as well as the diagram for how the ERC721 transfer works now


themacexpert (approved):
:thumbsup:

PR #74, opened 2021-04-23
Fix Connect MetaMask bugs

eshaben:
This PR includes fixes for the Connect MetaMask buttons...
- Fixes the bug when you disconnect via MetaMask and the buttons would still show "Connected"
- Better handles multiple addresses being connected
- Fixes another bug I found where the button would be disabled but if you double clicked it, it would still open MetaMask (because the event listener was still attached 🤦‍♀️ )


themacexpert (approved):
LGTM 👍

PR #75, opened 2021-04-28
Fix typos on send transaction, deploy contract, and voting page

themacexpert:
Fix typos on send transaction, deploy contract, and voting page


eshaben (approved):
LGTM ✨

PR #78, opened 2021-04-29
Add Precompiled Binary section for running nodes

wpank:
Adds a section on how to download and use precompiled binaries attached to releases.


JoshOrndorff:
Caution, these binaries have assumptions. For example they only work on x86-64 linux with certain versions of dependencies.

albertov19:
Yeah I've mentioned this over the collators channel. I'll just add a note in the docs saying "If you know what you are doing, you can directly download the precompiled binaries attached to releases (WITH URL)" Or something similar. With that note I'll close this PR but still super thankful for the suggestion.

PR #80, opened 2021-05-03
remove all instances of "we want to hear from you"

eshaben:
As part of the docs revamp, we're getting rid of all of the "we want to hear from you" sections in favor of a single section in the footer. This PR just removes all instances of "we want to hear from you".

This PR kind of goes hand in hand with [mkdocs PR #6 ](https://github.com/PureStake/moonbeam-mkdocs/pull/6) and can be merged really whenever but I recommend waiting until the mkdocs PR has been merged and released to production!


albertov19 (approved):
Bye bye "we want to hear from you"

PR #83, opened 2021-05-06
update docker commands in telemetry section

eshaben:
the `--chain alphanet` comand is listed on the moonbeam repo readme as required to join the testnet so just updating the telemetry section of the docs to reflect this change as well


albertov19:
This is only required to set the different ChainSpecs for Moonbeam.

PR #90, opened 2021-05-20
MD66- Revise Governance Tutorial

themacexpert:
MD66- Revise Governance Tutorial to change from a set balance function to a system remark. Update images to reflect new tutorial process and maintain same account names as before.


albertov19 (changes requested):
Looks good!!!

A minor nitpick ¯\_(ツ)_/¯ :

- When taking screenshots that include clicking on a button I tend to hover my mouse over the button so that the animation is displayed. You can take a look at the current Proposals site to see what I mean.

I would suggest as well to revisit the voting tutorials page because in there we are still making a reference to the old tutorial.

themacexpert:
> I would suggest as well to revisit the voting tutorials page because in there we are still making a reference to the old tutorial.
I revised the voting tutorials page to reflect the new system remark in text and images. 

> * When taking screenshots that include clicking on a button I tend to hover my mouse over the button so that the animation is displayed. You can take a look at the current Proposals site to see what I mean.
Noted, I started making screenshots this way for the new images on the voting page.

PR #94, opened 2021-05-27
Moonbase v8

albertov19:
This PR introduces the changes for the v8 update.


albertov19:
Will merge because of the Moonbase v8 release

PR #96, opened 2021-05-27
cleanup rpc flags

artkaseman:
The RPC flags were inconsistent, so I defaulted to no external RPC, and RPC methods = Auto (the default) which will be safe if externally expose and unsafe if not.  

We can't create a template that is good for everyone, so a default base secure template will be best, and we can help out RPC providers if they need additional help.

### DIFF
--- a/node-operators/networks/full-node.md
+++ b/node-operators/networks/full-node.md
@@ -116,8 +116,6 @@ Now, execute the docker run command. Note that you have to:
     docker run -p {{ networks.parachain.rpc }}:{{ networks.parachain.rpc }} -p {{ networks.parachain.ws }}:{{ networks.parachain.ws }} -v "{{ networks.moonbase.node_directory }}:/data" \
     -u $(id -u ${USER}):$(id -g ${USER}) \
     purestake/moonbeam:{{ networks.moonbase.parachain_docker_tag }} \
-    --ws-external \
-    --rpc-external \
     --base-path=/data \
     --chain alphanet \
     --name="YOUR-NODE-NAME" \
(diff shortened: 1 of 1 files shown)


node-operators/networks/full-node.md
```diff
@@ -119,7 +118,6 @@ Now, execute the docker run command. Note that you have to:
     purestake/moonbeam:{{ networks.moonbase.parachain_docker_tag }} \
     --ws-external \
     --rpc-external \
```
JoshOrndorff, line 120: You're recommending these for MacOS but not for Ubuntu.

node-operators/networks/full-node.md
```diff
@@ -159,7 +156,6 @@ Now, execute the docker run command. Note that you have to:
...
```
JoshOrndorff, line 158: Same for collators

node-operators/networks/full-node.md
```diff
@@ -293,10 +289,6 @@ The next step is to create the systemd configuration file. Note that you have to
          --ws-port {{ networks.parachain.ws }} \
          --pruning=archive \
          --state-cache-size 1 \
-         --unsafe-rpc-external \
-         --unsafe-ws-external \
-         --rpc-methods=Safe \
-         --rpc-cors all \
          --log rpc=info \
```
JoshOrndorff, line 292: This isn't a security concern, but why have collators log rpc stuff?

node-operators/networks/full-node.md
albertov19: Yes. My understanding is that network="host" doesn't work on mac so you have to provide these extra flags.

artkaseman: good point, fixed.

node-operators/networks/full-node.md
artkaseman: good point, fixed.

node-operators/networks/full-node.md
artkaseman: fixed

node-operators/networks/full-node.md
artkaseman: it was the --ws-external and --rpc external.  i left the host flags

PR #98, opened 2021-06-01
Faucet Page Updates

themacexpert:
A user in discord pointed out this weekend that our faucet page was out of date, so I have updated the page to include up to date screenshots as well as indicate the 5 DEV token limit and reminding users that DEV tokens have no value.


albertov19:
LGTM

PR #100, opened 2021-06-02
refactor language select

eshaben:
This PR includes an update to the language select to maintain the page the user is on when switching languages and a general cleanup. When I initially put the language select together, I was just trying to get something up quickly so it was a bit of a mess and there were some unneeded parts that I have removed in this PR.


albertov19:
Looks good 😎

albertov19 (approved):
Looks good 😎

PR #106, opened 2021-06-22
Hardware wallets

albertov19:
This PR adds two pages inside the /integrations/wallets section: Ledger and Trezor.


themacexpert (approved):
Looks good just two suggestions 👍

PR #107, opened 2021-06-23
add treasury overview documentation

eshaben:
Add an overview page on the one treasury that Moonbase Alpha has, the community treasury. There was initially a plan for two treasuries, however it was decided that the parachain bond treasury is not necessary as it does not require the full capabilities of a treasury. So instead the parachain bond treasury was turned into account set in staking storage. 

So this overview goes over the community treasury!


albertov19:
Looks gooooood!

Some comments:

 - Rememer to use variables from the `variables.yml` file. For example all the values in the table should be fetch from there, like `networks.moonbase.treasury.bond` and so on
 - I would make the tutorial Moonbeam generic, and just specify where certain things are Moonbase Alpha specific (this is something we might need to change in staking and governance as well). Now that we are approaching MainNet launch, we have to shift a little bit how we write documentation so it covers as many as all the networks as possible
 -

treasury/overview.md
```diff
@@ -0,0 +1,47 @@
...
+- **Proposal bond** - a deposit equal to a percentage of the total proposal spend amount
+- **Proposal bond minimum** - minimum amount for a proposal bond. This amount must be paid as the bond if it is higher than the deposit percentage
+- **Spend period** - the amount of days, in blocks, during which the treasury funds as many proposals as possible without exceeding the maximum
+- **Maximum approved proposals** - the maximum amount of proposals that can wait in the spending queue
+
+Currently, for Moonbase Alpha:
+
+|             Variable             |     |                                                  Value                                                  |
+| :------------------------------: | :-: | :-----------------------------------------------------------------------------------------------------: |
+|           Proposal bond          |     |                          5% of the proposed spend                                                       |
```
albertov19, line 29: Use the `variables.yml` file

treasury/overview.md
```diff
@@ -0,0 +1,47 @@
...
+|       Proposal bond minimum      |     |                          1 DEV token                                                                    |
+|           Spend period           |     |                          43200 blocks (6 days)                                                          |
+|     Maximum approved proposals   |     |                          100                                                                            |
+
+
+## Community Treasury
+
+To fund the Treasury, 20% of each block's transactions fees will be sent to the Treasury. The remaining 80% of the fees are burned. The Treasury allows stakeholders to submit spending proposals to be reviewed and voted on by the Council. These spending proposals should include initiatives to further the network or boost network engagement. Some network initiatives could include funding integrations or collaborations, community events, network outreach, and more.
```
albertov19, line 37: This first sentence sounds weird, probably "20% of each block's transactions fees will be allocated to it"? Or something similar

treasury/overview.md
```diff
@@ -0,0 +1,47 @@
...
+|     Maximum approved proposals   |     |                          100                                                                            |
+
+
+## Community Treasury
+
+To fund the Treasury, 20% of each block's transactions fees will be sent to the Treasury. The remaining 80% of the fees are burned. The Treasury allows stakeholders to submit spending proposals to be reviewed and voted on by the Council. These spending proposals should include initiatives to further the network or boost network engagement. Some network initiatives could include funding integrations or collaborations, community events, network outreach, and more. 
+
+To deter spam, proposals must be submitted with a deposit, also known as a proposal bond. If the proposal bond is less than the proposal bond minimum, the proposal bond minimum must be submitted for the deposit instead. So, any token holder that has enough tokens to cover the deposit can submit a proposal. If the proposer's balance doesn't have enough funds to cover the deposit, the extrinsic will fail due to insufficient funds and you will have to pay transaction fees.
```
albertov19, line 39: "If the proposal bond is less than the proposal bond minimum, the proposal bond minimum must be submitted for the deposit instead." -> "The proposal bond needs to be higher than a minimum amount, which can be changed by a governance proposal. So any token...."

albertov19, line 39: "If the proposer's balance doesn't have enough funds to cover the deposit, the extrinsic will fail due to insufficient funds and you will have to pay transaction fees."

"If the proposer doesn't have enough funds to cover the deposit ...... funds, but transaction fees will still be deducted."

albertov19 (changes requested):

eshaben:
ok this is ready to be reviewed again... I tried to make this more moonbeam-generic, I know we spoke about having different tables for the different values, but I figure we can cross that bridge if/when we need to since the values are the same across each network right now... as long as "1 Token" is ok with you, if not I can change to a table per network

eshaben:
this is good to be reviewed again

albertov19 (approved):

PR #113, opened 2021-07-01
Moonriver network page

eshaben:
This PR adds a Moonriver network page, similar to the page we have for [Moonbase Alpha](https://docs.moonbeam.network/networks/moonbase/).


variables.yml
```diff
@@ -107,7 +107,7 @@ networks:
     min_gas_price: 1
     block_time: 12
     gas_block: 15M
-    gas_tx: 13M
+    gas_tx: 15M
```
notlesh, line 110: Our block size is 15M gas, but there may be some little bits of overhead that prevent a transaction from consuming the entire thing (it might be the case that 14999999 fits, but not 15000000, etc.) If you need an exact number, I can get it for you.

networks/moonriver.md
```diff
@@ -18,23 +18,47 @@ In order to collect as much feedback as possible and provide fast issue resoluti
...
+
 | Variable              | Value                                       |
-|-----------------------|---------------------------------------------|
-| Minimum gas price     | {{ networks.moonriver.min_gas_price }} gsed |
-| Block time            | {{ networks.moonriver.block_time }} seconds |
-| Block gas limit       | {{ networks.moonriver.gas_block }}          |
-| Transaction gas limit | {{ networks.moonriver.gas_tx }}             |
-| RPC endpoint          | {{ networks.moonriver.rpc_url }}            |
-| WSS endpoint          | {{ networks.moonriver.wss_url }}            |
+|:---------------------:|:-------------------------------------------:|
```
eshaben, line 31: Thanks for fixing this! I noticed while working on the general Moonriver docs that all the tables needed to be centered

I shared a link in the developer relations channel for an extension that will make the tables prettier in case you're interested

networks/moonriver.md
```diff
@@ -18,23 +18,47 @@ In order to collect as much feedback as possible and provide fast issue resoluti
...
+| Minimum gas price     | {{ networks.moonriver.min_gas_price }} GSed ([token denominations](#token-denominations)) |
```
eshaben, line 32: I don't think I've ever really seen "Gwei" as "GWei"... I feel like this should be "Gsed" if anything but I don't know if it really matters

networks/moonriver.md
```diff
@@ -53,15 +77,15 @@ The tokens on Moonriver will also be called Moonriver (MOVR). Check out the Moon
...
-| Micromoonriver | 0.000001             | 1,000,000,000,000             |
-| Millimoonriver | 0.001                | 1,000,000,000,000,000         |
+| KiloSediment   | 0.000000000000001    | 1,000                         |
+| MegaSediment   | 0.000000000001       | 1,000,000                     |
+| GigaSediment   | 0.000000001          | 1,000,000,000                 |
+| MicroMoonriver | 0.000001             | 1,000,000,000,000             |
+| MilliMoonriver | 0.001                | 1,000,000,000,000,000         |
 | Moonriver      | 1                    | 1,000,000,000,000,000,000     |
-| Kilomoonriver  | 1,000                | 1,000,000,000,000,000,000,000 |
+| KiloMoonriver  | 1,000                | 1,000,000,000,000,000,000,000 |
```
eshaben: made a similar comment before about gwei but I don't know if I have ever seen "GigaWei" always "Gigawei"... just an observation but I don't know if it really matters

variables.yml
eshaben: @albertov19 do we need an exact number? could we just say "< 15M"

networks/moonriver.md
albertov19: Good point

networks/moonriver.md
albertov19: Yeah I changed this as well. Also, I think we'll use Kilosed instead if Kilosendiment (too long)

eshaben:
@albertov19 I think this is ready to be merged as long as you're good with where it stands now

networks/moonriver.md
```diff
@@ -0,0 +1,99 @@
...
+Some important variables to note include:
+
+### General Configurations
+
+|       Variable        |                                           Value                                           |
+|:---------------------:|:-----------------------------------------------------------------------------------------:|
+|   Minimum gas price   | {{ networks.moonriver.min_gas_price }} Gsed ([token denominations](#token-denominations)) |
+|      Block time       |          {{ networks.moonriver.block_time }} seconds (expected to be 6 seconds)           |
+|    Block gas limit    |         {{ networks.moonriver.gas_block }} (expected to increase by at least 4x)          |
+| Transaction gas limit |           {{ networks.moonriver.gas_tx }} (expected to increase by at least 4x)           |
```
albertov19, line 35: Tx gas limit is actually 11,874,999 haha it needs to be less than 11.875M

albertov19 (changes requested):

albertov19 (approved):

albertov19:
LGTM 🔥

PR #114, opened 2021-07-02
*Urgent* Updates for Collator Page to Include totalSelected Steps

themacexpert:
- Add get totalSelected steps 
- Add the candidatePoolSize step to join candidates steps
- Revise images


eshaben (commented):
Looks good to me. The only thing I would say is that the images should go in order and we're not using image 2 anymore. I think it goes 1 then 0 then 2a if you're reading through the document. So maybe just update it to be 0, 1, 2 🤷‍♀️ 

Not super important though, can be circled back to, as I see this is marked urgent

themacexpert:
Thanks Erin! Updated

PR #115, opened 2021-07-05
add french as supported language

albertov19:
Adds French as a supported language.

Also used standardized .js files with Prettier formatting


eshaben (approved):
LGTM ✨ 

do you have prettier config file we could all use?

PR #120, opened 2021-07-16
Nifty wallet

eshaben:
Create documentation around Nifty!
 
👉 Since all of the images are of a small screen, I had to use custom css to size the image so they wouldn't look too big on the docs page
👉 Nifty renders the token as ETH, there isn't a way to set it to MOVR. They do have a currency converter, but MOVR isn't on that list yet so left that as USD. Just included a note about it, but not sure if this is something we want to have fixed before publishing any docs as it could be pretty confusing 🤷‍♀️


integrations/wallets/nifty.md
```diff
@@ -0,0 +1,77 @@
...
+
+Nifty Wallet serves as a Web3 provider in tools such as [Remix](/integrations/remix/). By having Nifty Wallet connected to Moonbase Alpha, you can deploy contracts as you would like using MetaMask, signing the transactions with Nifty instead. 
+
+For example, in Remix, when deploying a smart contract, make sure you select the "Injected Web3" option in the "Environment" menu. If you have Nifty Wallet connected, you will see the TestNet chain ID just below the box (_1287_) and your Nifty Wallet account injected into Remix as well. When sending a transaction, you should see a similar pop-up from Nifty:
+
+<img src="/images/nifty/nifty-images-5.png" alt="Nifty sign transaction" style="width: 50%; display: block; margin-left: auto; margin-right: auto;" />
+
+Ensure you have DEV tokens in your account and, if necessary, head to the faucet to obtain some tokens. By clicking on "Submit," you are signing this transaction, and the contract will be deployed to the Moonbase Alpha TestNet.
+
+Please note that although your account balance shows ETH, it is not real ETH, and just DEV tokens.
```
albertov19, line 56: I would add a !!! note here.

albertov19 (changes requested):
Looks good! I suggested some changes

eshaben:
This is ready again!

integrations/wallets/nifty.md
```diff
@@ -0,0 +1,99 @@
...
+The transaction will be displayed under the "Sent" tab.
+
+<img src="/images/nifty/nifty-images-6.png" alt="Nifty confirmed transaction" style="width: 50%; display: block; margin-left: auto; margin-right: auto;" />
+
+## Create a New Account
+
+To create a new account, click the user icon in the top right corner and select "Create Account".
+
+<img src="/images/nifty/nifty-images-7.png" alt="Nifty create an account" style="width: 50%; display: block; margin-left: auto; margin-right: auto;" />
+
```
albertov19, line 89: I feel that not adding something here is kind of like an abrupt ending to this section. What happens after you click on "Create Account"?

integrations/wallets/nifty.md
```diff
@@ -0,0 +1,99 @@
...
+## Import an Account
+
+To create a new wallet, click the user icon in the top right corner and select "Import Account".
+
+<img src="/images/nifty/nifty-images-8.png" alt="Nifty import an account" style="width: 50%; display: block; margin-left: auto; margin-right: auto;" />
+
+Next, select the type of import from the dropdown and enter the necessary details to import your account. For example, if you choose to import with a private key, paste your private key in the input field and click "Import".
+
+<img src="/images/nifty/nifty-images-9.png" alt="MathWallet private key or mnemonic import" style="width: 50%; display: block; margin-left: auto; margin-right: auto;" />
+
```
albertov19, line 99: Same comment as before, what happens when you click on "Import"?

albertov19 (changes requested):
I've added some more comments, there are minor. I would also suggest that users can't connect Hardware wallets as it is not supported for Custom RPCs

PR #122, opened 2021-07-23
Fixes Broken Sectional Links in Setting Up A Node Page

hyd628:
Issue is related to mkdocs TOC extension slugifying headers that have been changed, creating a mismatch between the newly generated sectional IDs, and the existing URL's in the document. 

Ran across this while looking into a related issue in the Chinese docs: https://github.com/PureStake/moonbeam-docs-cn/pull/5


eshaben (approved):
Good catch! LGTM!

PR #125, opened 2021-07-30
Update staking precompile

eshaben:
Update the staking precompile tutorial to reflect recent changes to the `StakingInterface.sol` contract. Primarily adding in steps to get the collator nominator count and the nominator nomination count so that `nominate` can be called with all of the necessary parameters.

Updated screenshots as well.


themacexpert (changes requested):
Looks good we just need to change the nominatorCount method from this: 
api.query.parachainStaking.nominatorState(yourNominatorAccount);

to this: 
api.query.parachainStaking.nominatorState2(yourNominatorAccount);

Otherwise we will get an error in the console. Then ready to merge!

PR #133, opened 2021-08-31
Updating parachain_release_tag

calebcheng00:
Chain will not sync if on old node version v0.9.6


albertov19:
Thanks for taking the effort submit this PR! Community contributions are the best ❤️  - We had this change already in staging and it has already been published.

PR #134, opened 2021-08-31
remove duplicate min nomination variables

eshaben:
we had duplicate variables for min nomination, `min_nom_stake` and `min_nom_amount`, so just reduced that to `min_nom_amount`


albertov19:
Even though they seem to be the same thing they were different. One was the minimum amount you needed to have staked to be considered a nominator, and the second one was a minimum amount for each staking extrinsic (this latter one was removed I guess). We will change this in Henry's PR so there is no conflicts and is easier to merge.

Good catch!

PR #136, opened 2021-09-03
use iso format to display languages in select if user is on mobile

eshaben:
set the `isMobile` flag if user happens to be on a mobile device so the language select displays the iso formatted abbreviation of each language... 
Screen Shot 2021-09-03 at 2 55 46 PM


albertov19:
Checks passed ✔️ 🥳 !!

PR #138, opened 2021-09-03
update sed command

eshaben:


albertov19:
Yeah I think this should fix it!

PR #148, opened 2021-09-28
remove integrate metamask into a dapp article

eshaben:
This PR removes the integrate metamask into a dapp article and in the one place we were referencing the article, now it points users to the medium article 

https://medium.com/moonbeam-network/integrate-metamask-into-a-dapp-ea7528c5a786


albertov19:
Would it be possible to add a redirect to the medium article in the `mkdocs.yml`? It might be worth trying that

albertov19 (approved):
LGTM 🚀

PR #151, opened 2021-09-29
add hex chain values and clean up use of variables

eshaben:
This PR adds the chain id in hex anywhere the chain id is referenced. It also cleans up use of variables


albertov19 (approved):
I've fixed some minor issues, LGTM!! 🚀

PR #152, opened 2021-09-30
Fix command to get latest tag when compiling binary

gilmouta:
`git checkout tags/$(git tag | tail -1)` gets the last tag in alphabetical order.
Since 0.9.0 > 0.10.0 in alphabetical order, this isn't giving the latest version.
Correct command: `git checkout tags/$(git describe --tags)`


albertov19:
LGTM 🚀

PR #153, opened 2021-10-01
add index page icons and banner images for eth compare section

eshaben:
add index page icons and banner images for eth compare section


albertov19 (approved):
I like them!!! Great job 👍

PR #156, opened 2021-10-04
Update Remix Video to New version

themacexpert:
Update embedded remix Video to New version that also demos the Moonbeam Remix Plugin


albertov19:
LGTM!

PR #166, opened 2021-10-22
Collator Bond More and Collator Bond Less

themacexpert:
This PR includes updates to the collators page including:

- How to Bond Less as a Collator or Collator Candidate
- How to Bond More as a Collator or Collator Candidate
- Updates Current Max Collators in Moonbase Alpha and Moonriver


albertov19 (approved):
LGTM

PR #172, opened 2021-11-04
fix workflow

eshaben:
fix broken builds... we were already in the mkdocs dir when we tried to go back into it again


albertov19:
I'm closing this one as it seems it was a Github issue or something.

PR #176, opened 2021-11-16
update max Collators for MB Alpha and update release version tag

themacexpert:
- Update max Collators for Moonbase Alpha to 60.
- Update version release tag to 0.14.2

### DIFF
--- a/variables.yml
+++ b/variables.yml
@@ -127,7 +127,7 @@ networks:
     chain_id: 1285
     hex_chain_id: '0x505'
     node_directory: /var/lib/moonriver-data
-    parachain_release_tag: v0.13.0
+    parachain_release_tag: v0.14.2
     parachain_sha256sum: 0223dfeeec38a70f4fd5f4666944dfc5d3b6520c60f000a458f42121682fcfaf
     chain_spec: moonriver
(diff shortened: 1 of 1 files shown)


variables.yml
```diff
@@ -12,7 +12,7 @@ networks:
     hex_chain_id: '0x507'
     chain_spec: alphanet
     block_explorer: https://moonbase-blockscout.testnet.moonbeam.network/
-    parachain_release_tag: v0.13.2 # must be in this exact format for links to work
+    parachain_release_tag: v0.14.2 # must be in this exact format for links to work
     parachain_sha256sum: 9e748a0c91d5ed5f999bae91ceed1343db0d9dda40270201abf00db067c86a71
```
albertov19, line 16: Thanks for update the version, but you need to also update the `sha256_sum` of the new image.

albertov19 (changes requested):
Added a comment. You also need to update the release tag for Moonriver I guess, and the `sha_sum`

Thanks!

variables.yml
themacexpert: Thank you @albertov19! Fixed

albertov19 (approved):
LGTM

PR #178, opened 2021-11-17
Python Substrate Interface Finality Checking Sample code

hyd628:
Sample code and brief description for using Python Substrate Interface to check tx finality.


albertov19 (approved):
Looks good! This is great

PR #179, opened 2021-11-17
Fix up api providers

eshaben:
Separate API provider content so that the new API provider page only contains info for getting a custom endpoint and then put the chain-specific development and public endpoints in the corresponding get started > network pages


eshaben:
Ok @albertov19 I think this is good now 🤞 😅

albertov19 (approved):
LGTM!

PR #182, opened 2021-11-25
create erc20 precompile page and move staking precompile into new precompiles dir

eshaben:
This PR creates a new ERC20 precompile page, creates a new folder for Precompiled Contracts, and moves the Staking Precompile guide to the new directory.

Since images are changing there is a PR open in the chinese docs site, which will point to the new location of the images:  https://github.com/PureStake/moonbeam-docs-cn/pull/21

There are a couple redirects, so there is also a PR for the mkdocs repo: https://github.com/PureStake/moonbeam-mkdocs/pull/46


albertov19:
I pushed some changes, check them out and let me know what you think. From my side it is good to go

albertov19 (changes requested):
I think it looks good, the only thing we need to add is that this precompile does not include `deposit `and `withdraw` functions that are expected from a WETH (or similar) contracts. We will have a discussion in friday's learning session about this. We might add dummy functions that for deposit it will transfer the funds to himself and emit the event, and same for withdraw (but do not include this yet in the documentation)

eshaben:
ah yes! the note has been added. thanks!

albertov19 (approved):

PR #183, opened 2021-11-29
update erc-* and dapps for consistency

eshaben:
We were using ERC20 and ERC-20, so changed all references of ERC* to ERC-*

Noticed when I went to go update the [Moonbeam Docs Standards](https://purestake.atlassian.net/wiki/spaces/MOON/pages/1326383211/Moonbeam+Docs+Standards) that we also should be using DApps instead of dApps, so updated that too!


albertov19:
Great! Thanks!

albertov19 (approved):
Great thanks!

PR #190, opened 2021-12-06
Update Block Explorers on the Moonriver and Moonbase Alpha Network pages

themacexpert:
Quick PR to include the following:

- Update default block explorer variable to Moonscan (applicable such as when Adding a new Network in Metamask)
- Add Moonscan to the List of block explorers on the Moonbase Alpha and Moonriver network pages
- Add Collator Performance leaderboard
- Add DappLooker Dashboard Link


albertov19 (approved):
Good catch!

PR #199, opened 2021-12-13
results from using the compress images script

eshaben:
this is the result of using the new compress-images.py script... it compressed images larger than 900KB and overwrites existing image files ~~uses the compressed images on the docs site and the old larger images were renamed to "uncompressed-" + the image name~~


eshaben:
This goes with https://github.com/PureStake/moonbeam-mkdocs/pull/49

albertov19:
I haven't merged because of the reason that some images are bigger now haha. Let me know if all is good

eshaben:
I tried manually uploading and checking the images that show 100% and I still get the same response. I tried using the Pillow library to compress the image and it barely touches it. For example the `mathwallet-images-5.png` images original size is 1050.744140625KB and after compression using the Tinify library it grows to 1059.2841796875KB. When compressing the original image using Pillow the size still grows but not as much to 1052.5068359375KB.

Maybe something to do with the image that I am not understanding. I do know that Tinify library compresses the image by reducing the number of colors, and the images that are having trouble do seem to be predominantly white images... maybe that is the cause?

albertov19 (approved):
LGTM 🚀

PR #205, opened 2021-12-14
reorder network tabs

eshaben:
Orders tabs in the following order:

1. Moonbeam
2. Moonriver
3. Moonbase Alpha
4. Moonbeam Dev Node


albertov19 (changes requested):
I've made some edits. I fixed in instance that was missing and also standardized spacing because VSCode `.md` editor doesn't like the space between `=== "text-here"` and bullets lists or information below.

Please check :)

eshaben:
checked ✅

albertov19 (approved):

PR #210, opened 2021-12-22
Separate entries for nominators and delegators on glossary page

hyd628:
Update glossary.md to have separate entries for nominators and delegators on the glossary page.


albertov19 (approved):
👍

PR #216, opened 2021-12-28
create template for creating pull requests

eshaben:
This PR adds a new template for when we create pull requests so we can be reminded of all the little things we need to take into consideration such as not breaking the chinese site and compressing images


hyd628:
Thanks for making this. Should help keep things organized!

For these two parts:

- [ ] If images need to be deleted, I have made sure that the images can be deleted safely (doesn't impact the Chinese site) or I have created a Jira ticket to delete the images once translations are done.
- [ ] If variables need to be removed, I have either removed the variables safely (doesn't impact the Chinese site) or I have created a Jira ticket to delete the variables once translations are done.

I think if the information on what files/variables need to be cleaned up can be included in this template or somewhere else in the PR, the Jira tickets for deleting files/variables are not necessary as I will be able to just clean up the redundant files/variables as a part of the work on porting the PR.

hyd628 (approved):

themacexpert (approved):

PR #220, opened 2021-12-31
Fix VMP>DMP definition in overview.md

apocys:
### Description

Definition of DMP was incorrect based on https://wiki.polkadot.network/docs/learn-crosschain documentation

Only English version of the page exist - no need for any translation


eshaben:
Good catch! Thanks @apocys!

albertov19:
Great catch for sure, this was a mix up when passing it to the Moonbeam Docs Markdown format :)

PR #274, opened 2022-03-16
update pruning flag for the relay chain

eshaben:
### Description

Change the pruning flag for the relay chain from `archive` to `1000`

### Checklist

n/a

### Corresponding PRs

Goes with this chinese PR https://github.com/PureStake/moonbeam-docs-cn/pull/69

n/a

#### Items to be Updated

n/a


albertov19 (approved):
Minor comment, otherwise looks good!

PR #289, opened 2022-04-09
Fixed most links. However `Tools and Integrations` hyperlinks don't h…

x676f64:
Fixed most links. However `Tools and Integrations` hyperlinks don't have corresponding paths that I could fix.

### Description

I am correcting multiple broken links in the README.md file

### Checklist

- [ ] If images have been moved, I have created an additional PR in `moonbeam-docs-cn` to update images to their new paths

### Corresponding PRs

None that I am aware of

#### Items to be Updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,21 +27,21 @@ Moonbeam will also be a parachain on the Polkadot network. That means it will ge
 
 Currently, there are a few ways you can start building on Moonbeam: 
 
- - Build your own Moonbeam instance as a [development node](/builders/get-started/moonbeam-dev/)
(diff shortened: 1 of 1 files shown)


albertov19:
@x676f64 Thanks! Actually, before, the README use to be a page in our docs site. Now it is not anymore. @eshaben we should probably change the README so that it explains a bit the style of the documentation site so it is useful for anyone that wants to contribute.

We should also look into the hyperlinks that Taylor fixed to see if they need to be fixed elsewhere

eshaben:
Yes, thank you for fixing these and bringing this to our attention!

We don't have to worry about other links on the docs site because the redirects work on the published docs site. The readme is the only page that would not work with the redirects as its not published anymore.

PR #320, opened 2022-05-12
Update generating session keys section

eshaben:
### Description

In runtime 1600, new changes will be introduced that will impact how session keys are generated. Instead of calling `author_rotateKeys` twice to get a nimbus key and vrf key, it will only need to be called once and both keys will be returned at the same time.

Putting this in draft mode for now.

### Checklist

    - [x] After the script has been run, I have created an additional PR in `moonbeam-docs-cn`

### Corresponding PRs

Goes with this PR: https://github.com/PureStake/moonbeam-docs-cn/pull/118

#### Items to be Updated

These images will need to be removed: https://github.com/PureStake/moonbeam-docs/pull/320/commits/703711a403381ae6ed820630bcc97db355cb7d16


4meta5:
@eshaben https://github.com/PureStake/moonbeam/pull/1525 allows the user to directly copy paste the output from `author_rotateKeys` as input to AuthorMapping `register_keys` and `set_keys`.

4meta5:
That last PR didn't solve the problem.

https://github.com/PureStake/moonbeam/pull/1535 (in runtime 1600) makes the following changes:
1. correctly makes the input to `set_keys` a `Vec<u8>` so the user may directly copy paste the output from `author_rotateKeys` into the input of `set_keys`
2. removes `register_keys` so the user should call `set_keys` for the first time they set the keys as well as all subsequent times they rotate the keys
3. Deprecates `clear_association` and adds `remove_keys` to be used instead. `remove_keys` takes no inputs
4. changes the events associated with the extrinsics:
* For the first call to `set_keys`, the event is `KeysRegistered`
* For subsequent calls to `set_keys`, the event is `KeysRotated`
* For calls to `remove_keys`, the event is `KeysRemoved`

node-operators/networks/collators/account-management.md
```diff
@@ -78,20 +77,32 @@ To map your author ID to your account, you need to be inside the [candidate pool
...
 
-You can also check the current on-chain mappings by verifying the chain state. You'll need to click on **Developer** at the top of the page, then choose **Chain State** from the dropdown, and take the following steps:
+You can check the current on-chain mappings for a specific author ID or you can also check all of the mappings stored on-chain. For checking a specific author ID, you can take the first 64 hexadecimal characters of the concatenated public keys to get the author ID. To verify that the author ID is correct, you can run the following command with the first 64 characters passed into the `params` array:
+
+```
+curl http://127.0.0.1:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
+  "jsonrpc":"2.0",
+  "id":1,
+  "method":"author_hasKey",
+  "params": ["72c7ca7ef07941a3caeb520806576b52cb085f7577cc12cd36c2d64dbf73757a", "nmbs"]
```
albertov19, line 97: I find it confusing that here you have to slice the `0x` but in polkadot.js Apps you don't.

albertov19 (approved):
Looks good. Some minor feedback.

node-operators/networks/collators/account-management.md
eshaben: well you pass in "nmbs" to make sure it's the author id, you can also run `author_hasKey` and pass in "rand" as a param with the last 64 characters... but it's not really applicable in this section since checking the mappings just requires you pass in the author id/nimbus key

i can create a separate section that explains it?

PR #346, opened 2022-06-16
Added multisig API

jboetticher:
### Description

Added links & an example to the Moonbeam Safe to showcase its API

### Checklist

- [X] If pages have been moved around, I have created an additional PR in `moonbeam-mkdocs` to update redirects
- [X] If pages have been moved around, I have run the `move-pages.py` script to move the pages and update the image paths on the chinese repo
    - [X] After the script has been run, I have created an additional PR in `moonbeam-docs-cn`
- [X] If images have been added, I have run the `compress-images.py` script to compress the images.
- [X] If variables (in variables.yml) need to be updated (such as a name change), I have updated the `moonbeam-docs-cn` repo to use the new variables

### Corresponding PRs

#### Items to be Updated


tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
...
 Congratulations, you've successfully interacted with a smart contract using the Safe!
 
+## Using Moonbeam Safe APIs
+There are APIs available to read from and interact with Moonbeam Safes for Moonbase Alpha, Moonriver, and Moonbeam.
+
+|     Network    |                                           Multisig API Page                                              |
+|:--------------:|:--------------------------------------------------------------------------------------------------------:|
+|    Moonbeam    |  [{{ networks.moonbeam.multisig.api_page }}]({{ networks.moonbeam.multisig.api_page }}){target=_blank}   |
+|    Moonriver   | [{{ networks.moonriver.multisig.api_page }}]({{ networks.moonriver.multisig.api_page }}){target=_blank}  |
+| Moonbase Alpha |  [{{ networks.moonbase.multisig.api_page }}]({{ networks.moonbase.multisig.api_page }}){target=_blank}   |
```
eshaben, line 323: for something like this i would typically just use tabs since it's for each network, so it would be something like this: https://github.com/PureStake/moonbeam-docs/blame/master/builders/build/canonical-contracts/precompiles/democracy.md#L17-L30

tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
 
 Congratulations, you've successfully interacted with a smart contract using the Safe!
 
+## Using Moonbeam Safe APIs
```
eshaben, line 316:
```suggestion
## Using Moonbeam Safe APIs {: #using-moonbeam-safe-apis }

```

tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
...
+There are APIs available to read from and interact with Moonbeam Safes for Moonbase Alpha, Moonriver, and Moonbeam.
+
+|     Network    |                                           Multisig API Page                                              |
+|:--------------:|:--------------------------------------------------------------------------------------------------------:|
+|    Moonbeam    |  [{{ networks.moonbeam.multisig.api_page }}]({{ networks.moonbeam.multisig.api_page }}){target=_blank}   |
+|    Moonriver   | [{{ networks.moonriver.multisig.api_page }}]({{ networks.moonriver.multisig.api_page }}){target=_blank}  |
+| Moonbase Alpha |  [{{ networks.moonbase.multisig.api_page }}]({{ networks.moonbase.multisig.api_page }}){target=_blank}   |
+
+### Retrieve Safe Information
```
eshaben, line 325:
```suggestion
### Retrieve Safe Information {: #retrieve-safe-information }

```

tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
 
 Congratulations, you've successfully interacted with a smart contract using the Safe!
 
+## Using Moonbeam Safe APIs
+There are APIs available to read from and interact with Moonbeam Safes for Moonbase Alpha, Moonriver, and Moonbeam.
```
eshaben, line 317: always start with Moonbeam first, then Moonriver, and then Moonbase Alpha 🙂

tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
...
+|     Network    |                                           Multisig API Page                                              |
+|:--------------:|:--------------------------------------------------------------------------------------------------------:|
+|    Moonbeam    |  [{{ networks.moonbeam.multisig.api_page }}]({{ networks.moonbeam.multisig.api_page }}){target=_blank}   |
+|    Moonriver   | [{{ networks.moonriver.multisig.api_page }}]({{ networks.moonriver.multisig.api_page }}){target=_blank}  |
+| Moonbase Alpha |  [{{ networks.moonbase.multisig.api_page }}]({{ networks.moonbase.multisig.api_page }}){target=_blank}   |
+
+### Retrieve Safe Information
+You can retrieve information about Safes from the Moonbeam Safe API. From the Safe page, copy the address of your Safe:
+
+![Contract Interaction Call Result](/images/tokens/manage/multisig-safe/safe-30.png)
```
eshaben, line 328: for this image, there's actually a copy icon you can click on the left hand side panel i think pointing to that might be easier and then you don't have to write "copy" next to the arrow

tokens/manage/multisig-safe.md
```diff
@@ -312,4 +313,36 @@ To double check that the correct text was set, you can go through the process ag
...
+
+Now you can use the API:
+
+ 1. Open the corresponding network page from the Multisig API table above
+ 2. Scroll down to the **safes** section and click on the **/safes/{address}/** endpoint section to expand its panel
+ 3. Click the **Try it out** button on the right
+
+![Contract Interaction Call Result](/images/tokens/manage/multisig-safe/safe-31.png)
```
eshaben, line 336: There should be an arrow for all of the items, so I would add in the top of your chrome window so you can see the URL... and also so it's consistent with the other screenshots throughout the guide.

Also the arrow for item 2 isn't centered. I can schedule a quick call with you to go over some pointers for working with inkscape 🙂

tokens/manage/multisig-safe.md
eshaben: actually I don't know if this header is even necessary. I would probably just remove it and mention you can get all sorts of information about the safe, transactions, and more and this example will just show you how to use the API to get information about the safe

eshaben (changes requested):
The content is good! 👍 

Just some comments about formatting and the images. I'm sorry, I know I am a nitpicker 😬

tokens/manage/multisig-safe.md
albertov19: I would say that we should point at this being an example, but that there are many other methods available that you can check in each of the URLs?

jboetticher: Will remove the header & point to the fact that it's an example

tokens/manage/multisig-safe.md
jboetticher: The reason I told them to copy it was because the button copies with "mbase", so it would turn out "mbase:0x..."

We can try changing this in the actual website?

tokens/manage/multisig-safe.md
jboetticher: I've made the changes, and have uploaded the source files pngs, cropped pngs, and svgs to the google drive. 

Surprised that I didn't catch the 2 centered. Very ugly =(

tokens/manage/multisig-safe.md
eshaben: Ah I see... I really just wasn't a big fan of "copy" next to the arrow so I just removed it. The copy button actually does the same thing in the Rinkeby implementation where it has `rin:0x...` so I don't think we should change it here

eshaben (approved):
LGTM ✨

PR #371, opened 2022-08-04
Updated HRMP Values

albertov19:
We need to add this to the relevant translation ticket.

(cc @eshaben @hyd628)


eshaben (approved):
LGTM ✨

PR #380, opened 2022-08-17
manually set DPI and compress images

eshaben:
### Description

Ahrefs warned me that these two images (gelato-2.png & gelato-3.png) were too big... so manually resized and compressed them


albertov19 (approved):
✔️

PR #401, opened 2022-09-06
Eshaben/moonriver v0.26.0

eshaben:
### Description

Bumps Moonriver to v0.26.0

Goes with CN PR https://github.com/PureStake/moonbeam-docs-cn/pull/153

### DIFF
--- a/variables.yml
+++ b/variables.yml
@@ -239,9 +239,9 @@ networks:
     chain_id: 1285
     hex_chain_id: '0x505'
     node_directory: /var/lib/moonriver-data
-    parachain_release_tag: v0.25.0
-    parachain_sha256sum: a468d7f8fd834dda504d38138db8021573b0f0afbed3207957e773a6bf79072b
-    tracing_tag: purestake/moonbeam-tracing:v0.25.0-1701-e8cc
+    parachain_release_tag: v0.26.0
+    parachain_sha256sum: eff929c4ed6c5323cae16173eda048dcb647690c68ac06200d092e48de9b8655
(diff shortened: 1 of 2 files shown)


variables.yml
```diff
@@ -239,9 +239,9 @@ networks:
     chain_id: 1285
     hex_chain_id: '0x505'
     node_directory: /var/lib/moonriver-data
-    parachain_release_tag: v0.25.0
-    parachain_sha256sum: a468d7f8fd834dda504d38138db8021573b0f0afbed3207957e773a6bf79072b
-    tracing_tag: purestake/moonbeam-tracing:v0.25.0-1701-e8cc
+    parachain_release_tag: v0.26.0
+    parachain_sha256sum: eff929c4ed6c5323cae16173eda048dcb647690c68ac06200d092e48de9b8655
+    tracing_tag: purestake/moonbeam-tracing:sha-363761cd-opt-1800-a74b
```
albertov19, line 244: Why is this tracing image format different than those before?

albertov19 (approved):
Please check that the tracing image is correct :) - Otherwise it looks good

variables.yml
eshaben: It's the same one as moonbase alpha and the same one as mentioned here: https://purestake.slack.com/archives/C039A4CSQ2F/p1661526078993579

eshaben: same one as moonbase alpha and was mentioned in slack

albertov19: Let's do `sha-363761cd-opt-1800-latest` which is the same (same digest) but it has a more "standard" name 

https://hub.docker.com/r/purestake/moonbeam-tracing/tags

PR #403, opened 2022-09-06
remove deprecated staking fns

eshaben:
### Description

This PR (finally) removes the deprecated staking functions!

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/155


albertov19:
You know, after dealing with the the staking data, I think it might be good to leave them but just add that they were deprecated since runtime XYZ. What do you think?

eshaben:
@albertov19 Will they still be available in the API? Because they don't appear on Polkadot.js Apps as extrinsics anymore (Moonbase Alpha currently).... if they're not usable I don't see a point in keeping them 🤷‍♀️

albertov19:
> @albertov19 Will they still be available in the API? Because they don't appear on Polkadot.js Apps as extrinsics anymore (Moonbase Alpha currently).... if they're not usable I don't see a point in keeping them 🤷‍♀️

It will be available through the Polkadot.js SDK for example. Not through Apps (I believe)

eshaben:
ok i'll close this PR and mark the ticket for this as closed

PR #421, opened 2022-09-19
add collective precompile docs

eshaben:
### Description

This PR adds docs on the Collective precompile

### Checklist

### Corresponding PRs

n/a

#### Items to be Updated

n/a


albertov19 (changes requested):
Looks good! My only comment/doubt is why a proposal has to be approved first and then voted on by the collective. For example, is that the case for Treasury proposals? If so, our "happy path" in the treasury page does not reflect that.

I would add an explainer of around the different steps or something similar.

builders/pallets-precompiles/precompiles/collective.md
```diff
@@ -0,0 +1,208 @@
...
+2. Select the **proposals** extrinsic
+3. Toggle the **include option** slider off
+4. Click **+** to submit the query
+5. The results will appear below with the proposal index and the proposal details
+
+![View all treasury proposals](/images/builders/pallets-precompiles/precompiles/collective/collective-5.png)
+
+Now that you have the proposal and the proposal index, you'll be able to approve the proposal in the following section using the collective precompile.
+
+### Propose the Proposal {: #propose-te-proposal }
```
eshaben, line 142:
```suggestion
### Propose the Proposal {: #propose-the-proposal }
```

eshaben:
@albertov19 Olivia said the treasury page needs to be updated and she was looking at it similarly to the governance page. Not sure if we should just wait for the updates she is requesting or I can try to look at it and fix them but seems like duplicated efforts

What is described on the treasury page seems to reflect what I have here though:

```
Once a proposal has been submitted, the treasury council votes on it. The threshold for accepting a treasury proposal is at least three-fifths of the treasury council. On the other hand, the threshold for rejecting a proposal is at least one-half of the treasury council. Please note that there is no way for a user to revoke a treasury proposal after it has been submitted.

If approved by the treasury council, the proposal enters a queue to be placed into a spend period. If the spending queue happens to contain the number of maximum approved proposals, the proposal submission will fail similarly to how it would if the proposer's balance is too low. If the proposal gets rejected, the deposit will be lost and transferred to the parachain treasury account.
```

So I think the happy path diagram is correct unless I am missing something haha which could very well be the case

eshaben:
Changes will be made in a separate PR for treasury updates

albertov19 (approved):

PR #430, opened 2022-10-07
remove line breaks

eshaben:
### Description

Remove line breaks (<br>) from quick start page. They were mistakenly added and ended up causing 404s for the links they were wrapped around


themacexpert (approved):
Everything looks good! Not sure if you want to add a {target=_blank} to the Ledger link, that's the only thing I noticed.

eshaben:
good catch! thanks!

PR #432, opened 2022-10-11
add security considerations page

eshaben:
### Description

This PR adds docs around security considerations in the moonbeam vs ethereum section
### Checklist

### Corresponding PRs

N/A

#### Items to be Updated

N/A


builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,340 @@
...
+2. Click on **get**
+3. The response will appear right below the **get** button, which should be `hello`
+
+![Verify the whitelisted function was executed](/images/builders/get-started/eth-compare/security/security-8.png)
+
+To make sure that you cannot call any other function, you can repeat the previous steps to call the `whitelistFunctionSelector` function by taking the following steps:
+
+1. Expand the **whitelistFunctionSelector** function
+2. In the **_target** field, enter the contract address of the `SetMessage.sol` contract
+3. In the **_bytes** field, you can enter any other ABI-encoded call data. For example, you can use the bytes from the previous example that called
```
jboetticher, line 279: It might be useful to include an example of this in the same contract. For example:

```
contract SetMessage {
    string storedMessage;
    string nonWhitelistedMessage;

    function set(string calldata x) public {
        storedMessage = x;
    }

    function nonWhitelistedSet(string calldata x) public {
        nonWhitelistedMessage = x;
    }

    function get() public view returns (string memory) {
        return storedMessage;
    }
}
```

jboetticher (changes requested):
Left a few suggestions. I could get all of the examples working, so good job.

For content, I'm not sure if there are any other issues. I know there was an issue with the batch precompile, but I think it's still disabled.

On a side note, whenever selectors are brought up, I'm always reminded of the "attack" that uses selectors on proxy contracts. Though this is less of an attack and more so social engineering since it requires a user to believe that a contract isn't malicious due to its status as a proxy contract. Not very relevant, just here if anyone's interested in a "fun" fact:
https://medium.com/nomic-foundation-blog/malicious-backdoors-in-ethereum-proxies-62629adf3357

albertov19 (changes requested):
Looks good, left dome details. We should focus on showing where the problem lies, instead of the examples.

eshaben:
Ok this is ready now @albertov19 @jboetticher

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+As previously mentioned, one major concern of allowing low-level calls to arbitrarily execute code on Moonbeam is that Moonbeam has precompile contracts that can be called, which can be used to get around some protections that are typically available on Ethereum. To safely use arbitrary code execution on Moonbeam, you should consider the following, which **only applies to contracts that allow arbitrary code execution**:
+
+- Moonbeam [precompiled contracts](builders/pallets-precompiles/precompiles/){target=_blank} such as the Native ERC-20 precompile, XC-20 precompiles, and XCM-related precompiles allow users to manage and transfer assets without requiring access to the EVM. Instead, these actions are done using the Substrate API. So, if your contract holds native tokens or XC-20s and allows arbitrary code execution, these precompiles can be used to drain the balance of the contract
+- Setting a value of `0` when using the `call()` function (for example, `call{value: 0}(...)`) can be overridden by calling the native asset precompile and specifying an amount to transfer in the encoded call data
```
jboetticher, line 25: There's a small distinction that just occurred to me.

Let contract A be a contract with arbitrary code execution. Let contract B be a contract with a function that has:
`require(msg.value > 1 ether);`

Doing the following will still guarantee that the function on contract B cannot be called by contract A.
`call{value: 0}(...)`

Of course, the funds themselves are still at risk due to the precompile. Not sure if this is enough of an issue to really mention.

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+```
+function getBytes(address _contract, address _to) public view returns (bytes memory) {
+    // Load ERC-20 interface of contract
+    IERC20 erc20 = IERC20(_contract);
+    // Get amount to transfer
+    uint256 amount = erc20.balanceOf(address(this));
+    // Build the encoded call data
+    return abi.encodeWithSelector(IERC20.transfer.selector, _to, amount);
+}
+```
```
jboetticher, line 48: In the first draft, this was within the same contract that makes the arbitrary call. That's not apparent in this case, so the balanceOf call is confusing.

It might be better to just show an attack contract snippet (untested):


```
IERC20 public anyXC20Token;

function attack(address _arbitraryCallContract, address _beneficiary) public {
    // Get amount to transfer out of vulnerable contract
    uint256 amount = anyXC20Token.balanceOf(_arbitraryCallContract);

    // Build attack call data
    bytes attackCallData = abi.encodeWithSelector(IERC20.transfer.selector, _to, amount);

    // Do attack call
    _arbitraryCallContract.makeArbitraryCall(address(anyXC20Token), attackCallData);
}
```

jboetticher (changes requested):
The Precompiles Can Override a Set Value section has code that's a bit confusing in the way if you haven't read the first draft. Other than that, it looks good.

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Introduction {: #introduction }
+
+When developing smart contracts on Moonbeam, there are some security considerations to be aware of that do not apply when developing on Ethereum. Moonbeam has several [precompiled contracts](/builders/pallets-precompiles/precompiles/){target=_blank}, which are Solidity interfaces that enable developers to access Substrate-based functionality through the Ethereum API. Although the precompiled contracts are designed to improve the developer experience, there can be some unintended consequences that must be considered.
+
+This guide will outline and provide examples of some security considerations to be cognizant of when developing on Moonbeam.
+
+## Arbitrary Code Execution {: #arbitrary-code-execution }
+
+Smart contracts can allow arbitrary code exeuction through low-level message passing to other smart contracts via the [`call()` function](https://solidity-by-example.org/call/){target=_blank}, which is an available method for the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}.
```
albertov19, line 18:
```suggestion
Smart contracts can allow arbitrary code execution through low-level message passing to other smart contracts via the [`call()` function](https://solidity-by-example.org/call/){target=_blank}, which is an available method for the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}.
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+description: A description of the main differences that Ethereum developers need to understand in terms of security considerations when developing on Moonbeam.
+---
+
+# Security Considerations
+
+![Moonbeam v Ethereum - Security Banner](/images/builders/get-started/eth-compare/security-banner.png)
+
+## Introduction {: #introduction }
+
+When developing smart contracts on Moonbeam, there are some security considerations to be aware of that do not apply when developing on Ethereum. Moonbeam has several [precompiled contracts](/builders/pallets-precompiles/precompiles/){target=_blank}, which are Solidity interfaces that enable developers to access Substrate-based functionality through the Ethereum API. Although the precompiled contracts are designed to improve the developer experience, there can be some unintended consequences that must be considered.
```
albertov19, line 12:
```suggestion
When developing smart contracts on Moonbeam, there are some security considerations to be aware of that do not apply when developing on Ethereum. Moonbeam has several [precompiled contracts](/builders/pallets-precompiles/precompiles/){target=_blank}, which are Solidity interfaces that enable developers to access Substrate-based functionality through the Ethereum API, but circumventing the EVM. Although the precompiled contracts are designed to improve the developer experience, there can be some unintended consequences that must be considered.
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+As previously mentioned, one major concern of allowing low-level calls to arbitrarily execute code on Moonbeam is that Moonbeam has precompile contracts that can be called, which can be used to get around some protections that are typically available on Ethereum. To safely use arbitrary code execution on Moonbeam, you should consider the following, which **only applies to contracts that allow arbitrary code execution**:
+
+- Moonbeam [precompiled contracts](builders/pallets-precompiles/precompiles/){target=_blank} such as the Native ERC-20 precompile, XC-20 precompiles, and XCM-related precompiles allow users to manage and transfer assets without requiring access to the EVM. Instead, these actions are done using the Substrate API. So, if your contract holds native tokens or XC-20s and allows arbitrary code execution, these precompiles can be used to drain the balance of the contract
```
albertov19, line 24:
```suggestion
- Moonbeam [precompiled contracts](builders/pallets-precompiles/precompiles/){target=_blank} such as the Native ERC-20 precompile, XC-20 precompiles, and XCM-related precompiles allow users to manage and transfer assets without requiring access to the EVM. Instead, these actions are done using the Substrate API. So, if your contract holds native tokens or XC-20s and allows arbitrary code execution, these precompiles can be used to drain the balance of the contract, bypassing any security checks that are normally enforced by the EVM
```

builders/get-started/eth-compare/security.md
albertov19, line 25:
```suggestion
- Setting the value attribute of the transaction object to a fixed amount when using the `call()` function (for example, `call{value: 0}(...)`) can be bypassed by calling the native asset precompile and specifying an amount to transfer in the encoded call data
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+- Setting a value of `0` when using the `call()` function (for example, `call{value: 0}(...)`) can be overridden by calling the native asset precompile and specifying an amount to transfer in the encoded call data
+- Providing the function selector for the function that you want to allow to be executed is considered safe, assuming that the function call is safe and does not call precompiles
+- Blacklisting contracts that can execute arbitrary call data is not considered safe, as other precompiles might be added in the future. Providing whitelisted contracts that can execute the arbitrary call data is considered safe, assuming that the contracts being called are not precompiles, or that in the case they are, the contract making the call does not hold the native token or any XC-20
```
albertov19, line 27:
```suggestion
- Blacklisting target contracts (including precompiles) in the function that executes arbitrary call data is not considered safe, as other precompiles might be added in the future. Providing whitelisted target contracts in the function that executes the arbitrary call data is considered safe, assuming that the contracts being called are not precompiles, or that in the case they are, the contract making the call does not hold the native token or any XC-20
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+In the following sections, you'll learn about each of these security considerations through examples.
+
+### Precompiles Can Override a Set Value {: #setting-a-value }
+
+On Ethereum, a smart contract that allows for arbitrary code execution could force the value of a call to be `0`, guaranteeing that no amount of native currency would be sent with the transaction. Whereas on Moonbeam, setting the value of an arbitrary call to `0` can be overridden by calling the native ERC-20 precompile contract and transferring the native asset through the Substrate API. Since ERC-20s and XC-20s are not native assets, setting a value to `0` doesn't provide any protection for these types of assets on Ethereum or Moonbeam.
```
albertov19, line 33:
```suggestion
On Ethereum, a smart contract that allows for arbitrary code execution could force the value of a call to be a specific amount (for example, `{value: 0}`), guaranteeing that only that amount of native currency would be sent with the transaction. Whereas on Moonbeam, setting the value of an arbitrary call can be overridden by calling the native ERC-20 precompile contract and transferring the native asset through the Substrate API. Since ERC-20s and XC-20s are not native assets, setting the value attribute doesn't provide any protection for these types of assets on Ethereum or Moonbeam.
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+    // Arbitrary call
+    (bool success,) = _target.call(_bytes);
+    require(success);
+}
+```
+
+### Whitelisting Safe Contracts {: #whitelisting-safe-contracts}
+
+By whitelisting a specific contract address that can execute arbitrary call data, it ensures that the call is considered safe, assuming that the contracts being called are not precompiles. If they are precompiles, you'll want to make sure that the contract making the call does not hold the native token or any XC-20.
```
albertov19, line 89:
```suggestion
By whitelisting a specific target contract address in the function that can execute arbitrary call data, you can ensure that the call is considered safe, as the EVM will enforce that only whitelisted contracts can be called. This assumes that the contracts being called are not precompiles. If they are precompiles, you'll want to make sure that the contract making the call does not hold the native token or any XC-20. 
```

albertov19 (dismissed):
Looks good! Left some comments but I think it is ready to be reviewed by the devs :)

builders/get-started/eth-compare/security.md
eshaben: Ah yeah I see what you're saying. I'll make this function so it can be in any contract just so it's a bit more flexible

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+    (bool success,) = _target.call(_bytes);
+    require(success);
+}
+```
+
+### Whitelisting Safe Contracts {: #whitelisting-safe-contracts}
+
+By whitelisting a specific target contract address in the function that can execute arbitrary call data, you can ensure that the call is considered safe, as the EVM will enforce that only whitelisted contracts can be called. This assumes that the contracts being called are not precompiles. If they are precompiles, you'll want to make sure that the contract making the call does not hold the native token or any XC-20. 
+
+Blacklisting contracts from arbitrary code execution is not considered safe, as other precompiles might be added in the future.
```
notlesh, line 91: We use a range for our precompiles starting with low values, so it could be feasible to blacklist the entire we might use (e.g. `0-4096` or so. We should discuss internally before documenting this as a good technique, though...

notlesh (commented):
This looks good. I have a couple thoughts:

* The phrase `arbitrary code execution` is used in ways that aren't very clear to me. It might help to explain what exactly this means or to make a distinction between `arbitrary EVM execution` and an `arbitrary call`.

* Another difference between the standard Ethereum precompiles and our "custom" ones is the use of function selectors. The standard ones do not interpret their input data in the same way that call data is done, so they aren't callable. Ours do, which is why an EoA can call them or why we can `call()` them from another contract. I think this would be good to cover here.

eshaben:
> It might help to explain what exactly this means or to make a distinction between `arbitrary EVM execution` and an `arbitrary call`.

@notlesh What is the difference? Or what is an example of arbitrary EVM execution?

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,104 @@
...
+
+Smart contracts can allow arbitrary code execution through low-level message passing to other smart contracts via the [`call()` function](https://solidity-by-example.org/call/){target=_blank}, which is an available method for the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}.
+
+The `call()` function accepts ABI encoded call data that is then passed to the target contract. To get the encoded call data, you can use any of the [ABI encoding functions outlined in the Solidity docs](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#abi-encoding-and-decoding-functions){target=_blank}, including `abi.encodeWithSelector` which will be used in the examples in this guide. 
+
+As previously mentioned, one major concern of allowing low-level calls to arbitrarily execute code on Moonbeam is that Moonbeam has precompile contracts that can be called, which can be used to get around some protections that are typically available on Ethereum. To safely use arbitrary code execution on Moonbeam, you should consider the following, which **only applies to contracts that allow arbitrary code execution**:
```
notlesh, line 22: To illustrate my previous point:

`arb code execution` here is a bit confusing. From this context (this paragraph), it could mean:

* code executed from a smart contract in general -- from the EVM's perspective, this is arbitrary (other than following the normal rules)
* `allowing low-level calls to arbitrarily execute code on Moonbeam` is probably nothing different than other cases, but is a lot to swallow
* code that is executed as the target of some `call` (or similar) -- I think this is mainly what this article is concerned with

I think these are really all referring to the same thing, so maybe using (1) consistent phrasing and (2) a short intro giving context to "arb code exec" and esp. that we are referring to it as a result of some `*call` would help a lot.

eshaben:
@notlesh Applied feedback in my last commit. Def struggled a little bit to find better words to explain it so I may have missed the mark... LMK and thank you!! 🙌

builders/get-started/eth-compare/security.md
```diff
@@ -15,11 +15,13 @@ This guide will outline and provide examples of some security considerations to
 
 ## Arbitrary Code Execution {: #arbitrary-code-execution }
 
-Smart contracts can allow arbitrary code execution through low-level message passing to other smart contracts via the [`call()` function](https://solidity-by-example.org/call/){target=_blank}, which is an available method for the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}.
+Arbitrary code execution in Solidity is the ability to execute code and call functions of other contracts using an arbitrary number of arguments of any type.
```
notlesh, line 18: How about something like "A smart contract allows arbitrary execution of other contract code when it allows user-specified data when invoking a `call()`.

That probably needs a lot of work, but I think it includes some elements that you're trying to convey.

Specifically, I think we want to make it clear that we're talking about a smart contract making a call (as opposed to call from a EOA via a txn) in ways that are unintended.

notlesh: Here's the pattern we really care about (in its simplest form):

```
EOA signs transaction -> calls contract A -> calls contract B
```

This is problematic any time `contract A` allows user-input to influence its call to `contract B` -- this is true in Ethereum proper, of course, but we define the additional caveats that are introduced by Moonbeam here.

^ really just me summarizing the problem. What I think is missing in the discussion of "arbitrary code execution" is clarity on the pattern (specifically A -> B with some user-influence).

eshaben:
@albertov19 do you have any additional feedback based on recent changes?

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+
+This guide will outline and provide examples of some security considerations to be cognizant of when developing on Moonbeam.
+
+## Arbitrary Code Execution {: #arbitrary-code-execution }
+
+Arbitrary code execution in Solidity is the ability to execute code and call functions of other contracts using an arbitrary number of arguments of any type.
+
+A smart contract allows arbitrary execution of another contract when it allows a user to invoke the `call()` function and pass in arbitrary call data. The [`call()` function](https://solidity-by-example.org/call/){target=_blank} is made available through the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}. When the `call()` function is invoked, the target contract is called using the arbitrary call data.
+
+An example of arbitrary code execution, in it's simplest form, can be seen in the following diagram.
```
notlesh, line 22:
```suggestion
Arbitrary code execution follows the pattern in the diagram below when `Contract A` allows a user to influence its call to `Contract B`.
```

This might help make it clear what `Contract A` must do "wrong" in order to have potential issues.

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+
+This guide will outline and provide examples of some security considerations to be cognizant of when developing on Moonbeam.
+
+## Arbitrary Code Execution {: #arbitrary-code-execution }
+
+Arbitrary code execution in Solidity is the ability to execute code and call functions of other contracts using an arbitrary number of arguments of any type.
+
+A smart contract allows arbitrary execution of another contract when it allows a user to invoke the `call()` function and pass in arbitrary call data. The [`call()` function](https://solidity-by-example.org/call/){target=_blank} is made available through the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}. When the `call()` function is invoked, the target contract is called using the arbitrary call data.
```
notlesh, line 20:
```suggestion
A smart contract allows arbitrary execution of another contract when it allows a user to influence its own `call()` and pass in arbitrary call data and/or the `call()`s target. The [`call()` function](https://solidity-by-example.org/call/){target=_blank} is made available through the [address data type in Solidity](https://docs.soliditylang.org/en/latest/types.html#address){target=_blank}. When the `call()` function is invoked, the target contract is called using the arbitrary call data.
```

Disregard or change as you want, what I'm getting at is:

* Clarity on which call is affected (it's Contract A in our example, not the EOA transaction call)
* Point out that issues can arise when the target is arbitrary/influence-able, it doesn't just have to be the call data

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+
+An example of arbitrary code execution, in it's simplest form, can be seen in the following diagram.
+
+![Arbitrary code execution](/images/builders/get-started/eth-compare/security/security-1.png)
+
+As previously mentioned, one major concern of arbitrarily executing code on Moonbeam is that Moonbeam has precompile contracts that can be called, which can be used to get around some protections that are typically available on Ethereum. To safely use arbitrary code execution on Moonbeam, you should consider the following, which **only applies to contracts that allow arbitrary code execution**:
+
+- Moonbeam [precompiled contracts](builders/pallets-precompiles/precompiles/){target=_blank} such as the Native ERC-20 precompile, XC-20 precompiles, and XCM-related precompiles allow users to manage and transfer assets without requiring access to the EVM. Instead, these actions are done using the Substrate API. So, if your contract holds native tokens or XC-20s and allows arbitrary code execution, these precompiles can be used to drain the balance of the contract, bypassing any security checks that are normally enforced by the EVM
```
notlesh, line 28:
```suggestion
- Moonbeam [precompiled contracts](builders/pallets-precompiles/precompiles/){target=_blank} such as the Native ERC-20 precompile, XC-20 precompiles, and XCM-related precompiles allow users to manage and transfer assets without requiring access to the EVM. Instead, these actions are done using native Substrate code. So, if your contract holds native tokens or XC-20s and allows arbitrary code execution, these precompiles can be used to drain the balance of the contract, bypassing any security checks that are normally enforced by the EVM
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+- Setting the value attribute of the transaction object to a fixed amount when using the `call()` function (for example, `call{value: 0}(...)`) can be bypassed by calling the native asset precompile and specifying an amount to transfer in the encoded call data
+- Providing the function selector for the function that you want to allow to be executed is considered safe, assuming that the function call is safe and does not call precompiles
```
notlesh, line 30: Is this saying that the contract should provide the fn selector or that it's safe in some cases for the user to provide it? If the former, maybe say "hard coding the selector"?

Or instead of saying what's safe, say what isn't -- which I think is the combination of user-influence-able target (or any target that is our custom precompiles) and a fn selector.

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+- Blacklisting target contracts (including precompiles) in the function that executes arbitrary call data is not considered safe, as other precompiles might be added in the future. Providing whitelisted target contracts in the function that executes the arbitrary call data is considered safe, assuming that the contracts being called are not precompiles, or that in the case they are, the contract making the call does not hold the native token or any XC-20
```
notlesh, line 31:
```suggestion
- Blacklisting target contracts (including precompiles) in the function that executes arbitrary call data is **not** considered safe, as other precompiles might be added in the future. Providing whitelisted target contracts in the function that executes the arbitrary call data is considered safe, assuming that the contracts being called are not precompiles, or that in the case they are, the contract making the call does not hold the native token or any XC-20
```

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+
+In the following sections, you'll learn about each of these security considerations through examples.
+
+### Precompiles Can Override a Set Value {: #setting-a-value }
+
+On Ethereum, a smart contract that allows for arbitrary code execution could force the value of a call to be a specific amount (for example, `{value: 0}`), guaranteeing that only that amount of native currency would be sent with the transaction. Whereas on Moonbeam, setting the value of an arbitrary call can be overridden by calling the native ERC-20 precompile contract and transferring the native asset through the Substrate API. Since ERC-20s and XC-20s are not native assets, setting the value attribute doesn't provide any protection for these types of assets on Ethereum or Moonbeam.
```
notlesh, line 37: Could be a good place for a link to related documentation on our erc20 precompiles

albertov19:
> @albertov19 do you have any additional feedback based on recent changes?

I would address @notlesh comments, besides this all good on my end

builders/get-started/eth-compare/security.md
```diff
@@ -0,0 +1,108 @@
...
+
+```
+function makeArbitraryCall(address _target, bytes calldata _bytes) public {
+    // Value: 0 does not protect against native ERC-20 precompile calls or XCM precompiles
+    (bool success,) = _target.call{value: 0}(_bytes);
+    require(success);
+}
+```
+
+The value of `0` will be overridden by the amount to be transferred as specified in the encoded call data, which in this example is the balance of the contract.
```
notlesh, line 64: It might be worth mentioning that we have a way to use our native currency as an ERC20, whereas there is no such thing in Ethereum. In other words:

* Ethereum provides only `value` as a way to transfer funds
* Moonbeam provides `value` and an ERC20-compatible precompile to move the native token

I think this is the fundamental reason this potential security issue arises.

notlesh (approved):
Looks good!

jboetticher (approved):
LGTM

builders/get-started/eth-compare/security.md
albertov19: @chebii2 this is not the right place to look for support. Please contact the team via the official support channels (Telegram/Discord) which you can find in https://moonbeam.network

PR #436, opened 2022-10-13
Minor updates to some precompile docs

eshaben:
### Description

✅ Update the precompiles pages where the steps in the Remix Set Up sections don't align with the images so that they now do align
✅ Update "Injected Web3" to "Injected Provider - Metamask" as the Remix UI has been updated
✅ Update " ...provide the address of the precompile in the At Address Field" so that field is lower case

### Checklist

### Corresponding PRs

#### Items to be Updated

 ⭕️ Need to delete `call-1.png` and then rename `call-1-new.png` to `call-1.png`


albertov19 (approved):
LGTM

PR #449, opened 2022-10-25
update pruning flags

eshaben:
### Description

The pruning flag has been replaced with `--blocks-pruning` and `--state-pruning`. The default for `--blocks-pruning` is all blocks and the default for `--state-pruning` is the last 256 blocks.

This PR makes the following changes:

- Changes to collator flags:
  👉 we can use the default for `--state-pruning` and `--blocks-pruning`

- Changes to full nodes:
  👉 `--state-pruning` should be `archive` for the parachain flags and we can use the default for `--state-pruning` for the relay chain. We can also use the default for `--blocks-pruning`
  
Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/177

### Checklist

### Corresponding PRs

https://github.com/PureStake/moonbeam-docs-cn/pull/177

#### Items to be Updated

N/A

### DIFF
--- a/.snippets/text/full-node/macos-node.md
+++ b/.snippets/text/full-node/macos-node.md
@@ -15,11 +15,10 @@ purestake/moonbeam:v0.26.1 \
 --name="YOUR-NODE-NAME" \
 --execution wasm \
 --wasm-execution compiled \
---pruning archive \
+--state-pruning archive \
 --state-cache-size 0 \
 -- \
 --execution wasm \
---pruning 1000 \
 --name="YOUR-NODE-NAME (Embedded Relay)"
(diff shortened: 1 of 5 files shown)


albertov19 (approved):
Should we add a note that this is only with the latest client? The old client still uses the old flags

eshaben:
> Should we add a note that this is only with the latest client? The old client still uses the old flags

There are a lot of notes surrounding these commands which is why I initially opted out of adding any additional notes. And also because the commands are for runtime `v0.27.0` 🤷‍♀️

albertov19 (approved):
LGTM

PR #461, opened 2022-11-02
General Fixes

albertov19:
### Description

Slightly modifies the remote EVM calls diagram.

Changes the first XCM instruction for token transfers as it was not the right one. (@eshaben can you help with the translation of this last bit?)


eshaben (approved):
LGTM

created this CN PR to go with: https://github.com/PureStake/moonbeam-docs-cn/pull/181

PR #468, opened 2022-11-11
minor updates to batch precompile docs

eshaben:
### Description

This PR fixes the following:

👉 Use **message** instead of **m** (contract changed at some point)
👉 Cannot pass an empty string in as an argument, must be `0x` (reported by community member 🙌 )
👉 Link is supposed to be for the batch precompile, but it was for the proxy precompile... Oops! 😬 

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/186


themacexpert (approved):
LGTM 👍

PR #469, opened 2022-11-15
add new section on bypassing tx.origin==msg.sender check in moonbeam

eshaben:
### Description

Add new section to the Security Considerations page as the call permit and batch precompiles can be used to bypass the require(tx.origin == msg.sender) check allowing multiple contract interactions at once by an EOA, which cannot be done in Ethereum. In Ethereum, an EOA can only interact with a contract once per transaction.

### Checklist

### Corresponding PRs

n/a

#### Items to be Updated

n/a

### DIFF
--- a/builders/get-started/eth-compare/security.md
+++ b/builders/get-started/eth-compare/security.md
@@ -106,3 +106,24 @@ function makeArbitraryCall(address _target, bytes calldata _bytes) public {
     require(success);
 }
 ```
+
+## Bypassing `tx.origin == msg.sender` Requirement {: #bypassing-txorigin-msgsender-requirement }
+
+The transaction origin, or `tx.origin`, is the address of the externally owned account (EOA) the transaction originated from. Whereas the `msg.sender` is the address that has initiated the current call. The `msg.sender` can be an EOA or a contract. The two can be different values if one contract calls another contract, as opposed to directly calling a contract from an EOA. In this case, the `msg.sender` will be the calling contract and the `tx.origin` will be the EOA that initially called the calling contract.
+
+For example, if Alice calls a function in contract A that then calls a function in contract B, when looking at the call to contract B, the `tx.origin` is Alice and the `msg.sender` is contract A.
+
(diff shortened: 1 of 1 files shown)


eshaben:
@albertov19 any suggestions for the section name? I feel like the current one is not all that great. Was struggling to find something on the shorter side 😅

builders/get-started/eth-compare/security.md
```diff
@@ -105,4 +105,25 @@ function makeArbitraryCall(address _target, bytes calldata _bytes) public {
...
+You can use the [require function](https://docs.soliditylang.org/en/v0.8.17/control-structures.html#panic-via-assert-and-error-via-require){target=_blank} to compare the `tx.origin` and `msg.sender`. If they are the same address, you're ensuring that only EOAs can call the function. If the `msg.sender` is a contract address, an exception will be thrown.
+
+```
+function transferFunds(address payable _target) payable public {
+    require(tx.origin == msg.sender);
+    _target.call{value: msg.value};
+}
+```
+
+On Ethereum, you can use this check to ensure that a given contract function can only be called once by an EOA. This is because on Ethereum, EOAs can only interact with a contract once per transaction. However, on Moonbeam, **this check can be bypassed** as EOAs can interact with a contract multiple times at once through the [batch](/builders/pallets-precompiles/precompiles/batch){target=_blank} and [call permit](/builders/pallets-precompiles/precompiles/call-permit){target=_blank} precompiles.
```
albertov19, line 125:
```suggestion
On Ethereum, you can use this check to ensure that a given contract function can only be called once by an EOA. This is because on Ethereum, EOAs can only interact with a contract once per transaction. However, **this is not the case** on Moonbeam, as EOAs can interact with a contract multiple times at once through the [batch](/builders/pallets-precompiles/precompiles/batch){target=_blank} and [call permit](/builders/pallets-precompiles/precompiles/call-permit){target=_blank} precompiles.
```

albertov19 (changes requested):
Looks good, left some minor comments. Not sure about the section name tbh 😞

eshaben:
Tried to update the title of the section, at least it doesn't use back ticks, but not sure it's any better 😆

albertov19 (changes requested):
Precompiles can Bypass Sender vs Origin Checks?

The current one is very very generic

albertov19 (approved):

PR #472, opened 2022-11-18
add security note

eshaben:
### Description

Add note to precompile pages to refer back to security considerations page. Only included some of the precompiles, wasn't sure if it made sense to add them to all of the precompiles but lmk and I'll be happy to add them!

Once the text of the snippet is reviewed, I'll ask Yuki to translate it and create a PR for the CN repo.

### Checklist

### Corresponding PRs

n/a

#### Items to be Updated

n/a


albertov19 (approved):
We will only add this in this precompiles? Why not all? Just because we don't think there is any specific problems with the others?

eshaben:
I'm happy to add them in all just in case. But yeah I added it to the ones where I thought there would be problems. But I'm not a Solidity expert so I'll add it to all

PR #476, opened 2022-11-22
update bridges to gmp protocols

eshaben:
### Description

Update "Bridges" to "GMP Protocols"

### Checklist

- [x] If pages have been moved around, I have created an additional PR in `moonbeam-mkdocs` to update redirects
- [x] If pages have been moved around, I have run the `move-pages.py` script to move the pages and update the image paths on the chinese repo
    - [x] After the script has been run, I have created an additional PR in `moonbeam-docs-cn`

### Corresponding PRs

https://github.com/PureStake/moonbeam-docs-cn/pull/189
https://github.com/PureStake/moonbeam-mkdocs/pull/94


albertov19:
@jboetticher what do you think?

jboetticher:
No problem with calling it GMP. Might be better for branding if we called them connected contracts, since all of our marketing will be calling them that

albertov19:
> No problem with calling it GMP. Might be better for branding if we called them connected contracts, since all of our marketing will be calling them that

But connected contracts is like the strategy, not the GPM Protocols themselves. Wormhole, LayerZero and Hyperlane don't refer themselves as GMPs. We might also consider Interoperability?

jboetticher:
If you call it Interoperability, do you plan on adding other non-bridging/messaging related projects as well? 
I think `Connected Contracts` is the closest thing that we have that will apply to all the other protocols. Or maybe `Cross-Chain Messaging`

eshaben:
> We could add everything that is `Interoperability` related here, but I also like the `Cross-Chain Messaging` part (and it can still be called `gmp` for URLs? Or keep the `bridges` to avoid redirects? What do you think @eshaben ?

I don't have a huge preference between either `Interoperability` or `Cross-Chain Messaging`. Maybe leaning slightly towards `Cross-Chain Messaging`. So would we have `Cross-Chain Messaging` (or `Interoperability`) as the main dir then `XCM` and `GMP` as subdirs? This would also create more redirects, but I feel like having `XCM Interoperability` and `Interoperability` might seem confusing at first glance. Although our redirects list is quite long, I don't know if I would say we need to keep certain pages as is for that reason alone 🤷‍♀️

albertov19:
> > We could add everything that is `Interoperability` related here, but I also like the `Cross-Chain Messaging` part (and it can still be called `gmp` for URLs? Or keep the `bridges` to avoid redirects? What do you think @eshaben ?
> 
> I don't have a huge preference between either `Interoperability` or `Cross-Chain Messaging`. Maybe leaning slightly towards `Cross-Chain Messaging`. So would we have `Cross-Chain Messaging` (or `Interoperability`) as the main dir then `XCM` and `GMP` as subdirs? This would also create more redirects, but I feel like having `XCM Interoperability` and `Interoperability` might seem confusing at first glance. Although our redirects list is quite long, I don't know if I would say we need to keep certain pages as is for that reason alone 🤷‍♀️

I think we should go with `Cross-Chain Messaging` and have XCM listed there but that will take you to the XCM Overview page that we have somewhere else.

The new name for this section could be `xchain`? or if you don't want redirects we could keep it as `bridge` but if we are going to expand this section we might as well rename it now

eshaben:
> The new name for this section could be `xchain`? or if you don't want redirects we could keep it as `bridge` but if we are going to expand this section we might as well rename it now

I don't think it makes sense to link to the XCM overview page because then on the side bar it would still be `XCM Interoperability` and `Cross-Chain Messaging` and I really think XCM should be moved to under `Cross-Chain Messaging` so it's like:

- `Cross-Chain Messaging`
   - `XCM`
     - all the xcm stuff
   - `GMP`
     - axelar plus all the new gmp docs coming 

Are you proposing that it's like:

- `XCM Interoperability`
  - all the XCM stuff
- `Cross-Chain Messaging`
   - `XCM` > points to XCM overview 
   - `X-Chain` (for GMP stuff?)

I personally think we can rule out `bridge`.

jboetticher:
As Alberto mentioned before, `GMP` is really just a terminology that Axelar has adopted. Wormhole says `xChain` and LayerZero has `Omnichain`. I would say `Connected Contracts` since that's our own terminology, but I'm unsure if that's supposed to also cover XCM

eshaben:
closing this, will make changes in a new PR

PR #495, opened 2022-12-21
fix 404 and image that is too large

eshaben:
### Description

Ahrefs report came back with a 404 and an image too large error. So this PR fixes both of those things. It also fixes a spacing issue in the checking prereqs section of the eth and moonbeam ledger pages


albertov19 (approved):
LGTM

PR #498, opened 2022-12-23
add tutorial on batching approval and swaps

eshaben:
### Description

This PR adds a new tutorial for batching approval and swaps into a single transaction with the `batchAll()` function of the batch precompile.

I'll get to the images and checklist next week.

### Checklist

### Corresponding PRs

#### Items to be Updated


tutorials/batch-approve-swap.md
```diff
@@ -0,0 +1,310 @@
...
+  const batch = await ethers.getContractAt("Batch", batchAddress);
+}
+main();
+```
+
+### Add Check Balances Helper Function {: #add-function-to-check-balances }
+
+Next, we're going to create a helper function that will be used to check the balance of DTOK tokens the DEX and the signer account has. This will be particularly useful to see balance changes after the swaps are complete.
+
+In this function, you'll call the `balanceOf()` function of the `DemoToken` contract, passing in the address of the signer and the DEX, and then print the formatted results in DTOKs to the terminal:
```
albertov19, line 207: I would state that the contract has an ERC-20 interface so you can check the balance using the normal `balanceOf` function

tutorials/batch-approve-swap.md
```diff
@@ -0,0 +1,310 @@
...
+After the swap, we'll check the balances again using the `checkBalances()` function to make sure the balances have changed as expected.
+
+We'll update the `main()` function to include the following logic:
+
+```js
+async function main() {
+  // ...
+
+  // Parse the value to swap to Wei
+  const amountDtok = ethers.utils.parseEther( "INSERT-AMOUNT-OF-DEV-TO-SWAP")
```
albertov19, line 280: `INSERT-AMOUNT-OF-DTOK-TO-SWAP` right?

albertov19 (changes requested):
Looks good! I was expecting you doing this but against the Uniswap V2 demo app we have in Moonbase, but this is good as well. 

If you want, you can create a section at the end where you talk about how the process would look like when implementing this against an Uniswap V2 styled dex. The most important thin is understanding how to build the encoded call data of [this function](https://github.com/PureStake/moonbeam-uniswap/blob/6e3ecf90fecb77bdea84d320fbaae6512f5e0e69/uniswap-contracts-moonbeam/contracts/periphery/UniswapV2Router02.sol#L284). Which is:

- Amount in
- Amount out
- Path array (in this case it would be [DTOK, ERC-20 PRECOMPILE] I think (or ERC-20 address if you want to use `MARS` for example)
- Recipient Address
- Deadline (TIMESTAMP)

Try it against Moonbeam-Swap first to ensure that it works

eshaben:
Thanks for reviewing. I didn't realize that you were thinking about uniswap. Figured I would use a simple example, but I incorporated your feedback and added a section on using the batch logic with uniswap. Just used the existing tokens set up on Moonbeam-swap and to relate back to previous example where we were swapping an ERC-20 for the native asset, I decided to swap ERTH for DEV. Tested it out and it works as expected.

tutorials/batch-approve-swap.md
```diff
@@ -0,0 +1,419 @@
...
+  const signerBalance = ethers.utils.formatEther(
+    await demoToken.balanceOf(signer)
+  );
+  console.log(
+    `Account ${signer} has a balance of: ${signerBalance} DTOKs`
+  );
+}
+```
+
+### Swap DEV for Tokens {: #add-logic-to-swap-dev }
```
jboetticher, line 246: My initial thought was to ax this section and just have the smart contract directly mint to the deployer's contract, since this section feels like setup that should be removed if possible so that the reader can get to the point of the tutorial faster.

```solidity
contract DemoToken is ERC20 {
    constructor(uint256 initialSupply) ERC20("DemoToken", "DTOK") {
        _mint(msg.sender, initialSupply / 2);
        _mint(tx.origin, initialSupply / 2);
    }
}
contract SimpleDex {
    // Made this payable so that liquidity can be added with deployment
    constructor() payable {
        // Mint 1000 DTOK tokens
        token = new DemoToken(1000000000000000000000);
    }
   //...
}
```

But it also looks like one of the main reasons that this step is needed is so that DEV liquidity exists for the contract. This can also be done in the constructor. If you choose not to remove this section, it would be good to also explain that liquidity is being added for the batch swap

jboetticher (commented):
Looks good 👍 . The only substantial change that I suggested was to consider replacing the first swap part with constructor logic so that the reader can get to the point of the tutorial faster.

tutorials/batch-approve-swap.md
eshaben: Yeah I agree that it's probably easiest to just do this in the constructor. Thanks for bringing this up!

tutorials/batch-approve-swap.md
```diff
@@ -0,0 +1,402 @@
...
+```
+mkdir scripts && touch scripts/deploy.js
+```
+
+In the `deploy.js` script, you can paste in the following code, which will deploy the `SimpleDex` contract and print the address of the contract to the terminal upon successful deployment:
+
+```js
+async function main() {
+  // DEV liquidity to add
+  const value = ethers.utils.parseEther("INSERT-AMOUNT-OF-DEV");
```
jboetticher, line 158: Potentially add a suggested value?

jboetticher (approved):
👍

tutorials/batch-approve-swap.md
eshaben: I added a suggested value in the comment

eshaben:
Thanks for the reviews!

PR #502, opened 2023-01-06
fix broken links and update xcm fee variables

eshaben:
### Description

Fix broken links found from running an integrity scan. While fixing the broken links, I realized that the XCM fees page needed to be updated as the extrinsic base weight had changed for Polkadot and Kusama in November.

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/204


albertov19 (approved):
Thank you for making these changes!

PR #507, opened 2023-01-11
update the Randomness precompile docs

eshaben:
### Description

This PR adds information about editing the gas limit in MetaMask so that internal transactions have enough gas to successfully go through. There were also some changes to the Randomness and RandomnessConsumer contracts that required updating the import in order to compile the RandomnessLotteryDemo contract without errors.

### Checklist

### Corresponding PRs

N/A

- [x] Will need to create PR in `moonbeam-docs` repo to remove variables

#### Items to be Updated

images/builders/pallets-precompiles/precompiles/randomness/randomness-6.png - needs to be deleted
images/builders/pallets-precompiles/precompiles/randomness/randomness-6-new.png - needs to be renamed to randomness-6.png


albertov19 (changes requested):
The issue here is that the original sender who submits the request need to have provided enough gas so that the internal request can be executed successfully. There might be an issue with this check. We can circle back to this when discussed

albertov19 (approved):
Sorry I meant to approve it

PR #538, opened 2023-02-07
update ethers to v6

eshaben:
### Description

This PR updates our ethers-related docs to v6 (with the exception of hardhat, as ethers comes out of the box with hardhat and they haven't released v6 support yet). It also moves some of the code in the call permit precompile to new code snippets.

Please refer to [Ethers V6 migration guide](https://docs.ethers.org/v6/migrating/) for more info.

For the purposes of the docs site, the following changes were made:
- Update the provider from `new ethers.providers.StaticJsonRpcProvider` to [`new ethers.JsonRpcProvider`](https://docs.ethers.org/v6/api/providers/jsonrpc/)
- Use [`await contract.waitForDeployment();`](https://docs.ethers.org/v6/api/contract/#BaseContract-waitForDeployment) for contract deployments instead of `await contract.deployed()`
- Deployed contract addresses can now be retrieved from [`contract.target`](https://docs.ethers.org/v6/api/contract/#BaseContract-target) instead of `contract.address`
- To get contract deployed bytecode, we can now use [`await contract.getDeployedCode()`](https://docs.ethers.org/v6/api/contract/#BaseContract-getDeployedCode)
- Util methods are no longer accessed via `ethers.utils.*`, they can directly be access using `ethers.*` ([see related migration guide odds and ends section](https://docs.ethers.org/v6/migrating/#migrate-other))
- `ethers.utils.splitSignature` no longer exists, so now we can work with signatures through the [`ethers.Signature` class](https://docs.ethers.org/v6/api/crypto/#Signature)
- We can now access the `defaultAbiCoder` through the [`AbiCoder` class](https://docs.ethers.org/v6/api/abi/abi-coder/#AbiCoder)

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/221


albertov19 (approved):
I made one tiny edit, otherwise it looks good!

PR #570, opened 2023-03-06
hide feedback widget on index pages

eshaben:
### Description

Hides the feedback widget on index pages so we are only collecting feedback on page with content.

Goes with mkdocs PR: https://github.com/PureStake/moonbeam-mkdocs/pull/108


albertov19 (approved):
LGTM

PR #583, opened 2023-03-13
Update Polkadot.js API snippets

eshaben:
### Description

Update the Polkadot.js API snippets to use big numbers and disconnect the API after use
Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/243


albertov19 (approved):
LGTM

PR #587, opened 2023-03-14
J2 wormhole relayer update

jboetticher:
### Description

Wormhole's relayer engine project has matured a bit since the documentation has been written, as I found out late last week. Most of the content is the same, but I've archived my previous repository and replaced it with a new one that uses the relayer engine as a package instead of copying it in its entirety.

### Checklist

- [X] If images have been added, I have run the `compress-images.py` script to compress the images.
- [X] If this page requires a disclaimer, I have added one

#### Items to be Updated


builders/interoperability/protocols/wormhole.md
```diff
@@ -290,80 +294,79 @@ If you are using a chain that wasn’t listed in the EVM TestNet table above, yo
 ]
 ```
 
-
-Now that the wallets are sorted out for the executor, look at the code of the executor itself, which is in the `example-project/plugins/simplegeneralmessage_plugin/src/plugin.ts` file. If you haven’t been following along, the entire file can be accessed in [its GitHub repository](https://github.com/jboetticher/relayer-engine/blob/main/example-project/plugins/simplegeneralmessage_plugin/src/plugin.ts){target=_blank}.
+Now that the wallets are sorted out for the executor, look at the code of the executor itself, which is in the `example-project/plugins/simplegeneralmessage_plugin/src/plugin.ts` file. If you haven’t been following along, the entire file can be accessed in [its GitHub repository](https://github.com/jboetticher/relayer-engine-docs-example/blob/main/plugins/simplegeneralmessage_plugin/src/plugin.ts){target=_blank}.
```
eshaben, line 297: should the `example-project` part be dropped so its just `plugins/simplegeneralmessage_plugin/src/plugin.ts`

eshaben: if it does need to be dropped, please double check throughout the doc for other places it may be used

eshaben (changes requested):

eshaben (approved):
LGTM

PR #591, opened 2023-03-17
update polkadot.js app links

eshaben:
### Description

Update links to Polkadot.js Apps. There were a few still using the onfinality endpoint and there were a bunch that had `rpc=wss%3A%2F%2Fwss` in the link, so I updated them to be cleaner and use `rpc=wss://wss`

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/249


albertov19 (approved):
LGTM

PR #594, opened 2023-03-21
minor fixes

eshaben:
### Description

This PR fixes a few issues that Xinxin discovered:
- The `incrementer.increment` value in the ethers guide shouldn't be an array - the value should be passed in as a number. The test suite already uses a number as the value (https://github.com/PureStake/moonbeam-docs-test-suite/blob/main/test/builders/build/eth-api/libraries/ethers/deploy-contract.js#L134)
- A couple issues with variables on the governance pages

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/251


albertov19 (approved):
LGTM

PR #597, opened 2023-03-23
add script to check updated and created dates

eshaben:
### Description

This adds a script to fix the created date issue outlined in this mkdocs PR: https://github.com/PureStake/moonbeam-mkdocs/pull/111

There is an issue with a handful of pages where the plugin that adds the dates on each page cannot find the date a page was created and it automatically defaults to the current date. So, this script checks the dates on each page and if the created date is after the updated date, it renders the updated date as the created date instead.


albertov19 (approved):
LGTM

PR #608, opened 2023-04-05
Add a call permit tutorial for gasless transactions

eshaben:
### Description

This PR adds a new tutorial on using the call permit precompile to implement gasless transactions using DPS as an example.

### Checklist


tutorials/eth-api/call-permit-gasless-txs.md
```diff
@@ -0,0 +1,442 @@
...
+      chainId: {{ networks.development.chain_id }},
+      verifyingContract: '{{ networks.moonbase.precompiles.call_permit}}',
+    };
+    ```
+
+### Define the Typed Data Structure {: #define-typed-data-structure }
+
+Next, we'll need to define the typed data structure. The typed data structure defines the acceptable types of data that our users will be signing. If you take a look at the [`dispatch` function of the Call Permit Precompile](/builders/pallets-precompiles/precompiles/call-permit/#the-call-permit-interface){target=_blank}, you'll see the data that we need to send, along with the associated types, is as follows:
+
+```sol
```
albertov19, line 212: Maybe add a bullet list of what each input means? not sure what do you think?

Update: I noticed you explain them after, but I would add a note saying that you do :)

tutorials/eth-api/call-permit-gasless-txs.md
```diff
@@ -0,0 +1,442 @@
...
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+) external returns (bytes memory output);
+```
+
+We'll need to add each of the above parameters to our typed data structure, with a couple of modifications. We don't need to include the signature-related parameters, but we do need to include a parameter for the `nonce` of the `from` account, which will be a *uint256*. The signature-related parameters aren't needed at this point because we're building the message data for the users to sign. We'll circle back to the signature-related parameters after we've finished building the message and requested the signature.
+
+So, if we grab the rest of the parameters, we can start to build our data structure. The primary type will be `CallPermit` and it will be an array of objects that correspond to each of the parameters and define the `name` and `type` of each parameters:
```
albertov19, line 228: What do you mean with `The primary type will be "CallPermit"`?

tutorials/eth-api/call-permit-gasless-txs.md
```diff
@@ -0,0 +1,442 @@
...
+  ethersSignature.v,
+  ethersSignature.r,
+  ethersSignature.s,
+);
+
+await dispatch.wait();
+console.log(`Transaction hash: ${dispatch.hash}`);
+```
+
+??? code "View the complete script"
```
albertov19, line 431: I like this! But maybe put some styling that goes with Moonbeam

albertov19 (changes requested):
Splendid teachin', matey! I've scrawled some words fer ye. In general, I'd fancy more illustrations, so a chart or two'd be grand.

Additionally, methinks we be needin' a true workin' sample with a transaction hash and a course to it.

Nonetheless, this be shipshape, arr! (credit to ChatGPT hahaha)

tutorials/eth-api/call-permit-gasless-txs.md
eshaben: yeah this section is primarily about the data types and the following actually goes into the actual data and values being passed in. I'll add a note!

albertov19 (approved):
Looks good! Great diagrams and explanations :) 🚢

PR #611, opened 2023-04-07
minor updates to the send and execute xcm messages page

eshaben:
### Description

Was using the wrong calldata so this PR fixes that and applies some other minor fixes such as using `const` instead of `let` and using snake case instead of camel in python snippets

### Checklist


albertov19 (approved):
Lgtm

PR #624, opened 2023-04-19
fix dates

eshaben:
### Description

Fix dates on the tutorials, we're not in 2022 anymore 😂 

And remove dates from a few of the network pages since the dates are now on the bottom of the page


albertov19 (approved):
Good point :)

PR #625, opened 2023-04-19
minor change to running a tracing node steps

eshaben:
### Description

As reported in discord, update the tracing node steps so that the 3rd step doesn't seem like an optional step

Asked @wuxinxin2000 if translations are needed for this


albertov19 (approved):
LGTM

PR #631, opened 2023-04-20
Update XCM Fees for Moonbeam v0.31.1 and Kusama v0.9.38

eshaben:
### Description

Update XCM Fees for Moonbeam v0.31.1 and Kusama v0.9.38
This PR also includes variables for the client version of Polkadot and Kusama so we don't have to worry about opening a PR on the CN site when we change the version of the links (unless the specific line we've linked to has moved)

Goes with CN PR: https://github.com/PureStake/moonbeam-docs-cn/pull/272


albertov19 (approved):
LGTM! I do worry about this because of Weights V2 in which now the fee also depends on two values. We should sync if it makes sense to revamp this or not (depends on how impactful this page is)

PR #640, opened 2023-04-28
Modified External xc-20s

albertov19:
OpenGov can't submit `setCode` with `GeneralAdmin` so proposals doing this will fail

@eshaben can you add this PR to the changes in the translations from Local XC-20s plz


eshaben (approved):
LGTM! ✨

PR #656, opened 2023-05-09
update consensus and finality docs for eth 2.0

eshaben:
### Description

This PR updates the consensus & finality page for Etheruem now that its using PoS.

### Checklist

### DIFF
--- a/builders/get-started/eth-compare/consensus-finality.md
+++ b/builders/get-started/eth-compare/consensus-finality.md
@@ -95,3 +95,5 @@ You can find more information about Polkadot.js and the Substrate JSON RPC in th
 
 === "py-substrate-interface"
     --8<-- 'code/vs-ethereum/pysubstrateinterface.md'
+
+--8<-- 'text/disclaimers/third-party-content.md'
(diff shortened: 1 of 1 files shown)


builders/get-started/eth-compare/consensus-finality.md
```diff
@@ -13,48 +13,56 @@ While Moonbeam strives to be compatible with Ethereum's Web3 API and EVM, there
...
+
+In terms of finality, blocks on Ethereum can take quite a bit longer to finalize than on Moonbeam due to the checkpoint finality system it uses. In Ethereum, validators determine finality at checkpoint blocks, which are always the first block in an epoch. Since an epoch has 32 slots and each slot is 12 seconds, it'll take at least 384 seconds, or 6.4 minutes for a block to be finalized.
+
+Moonbeam does not use checkpoint blocks and instead relies on the GRANDPA finality gadget, where the finality process is completed in parallel to block production. In addition, the finality process incorporates the blockchain's structure, which allows the relay chain validators to vote on the highest block that they think is valid. In this scenario, the vote would apply to all of the blocks leading up to the one that is finalized, which speeds up the finalization process. After a block has been included in the relay chain, a block can be finalized within one block on Moonbeam.
+
+## Strategy to Check Transaction Finality {: #strategy-to-check-tx-finality }
```
albertov19, line 44: Ethereum introduced some new block tags like `finalized` etc as part of the PoS merge. We now support the same block tags in the Frontier RPC, so maybe there is another strategy here that we can show? Ethers.js and other libraries might support the block tags

albertov19 (changes requested):
Thanks for this! Left some feedback

albertov19 (approved):
We could also mentioned the Custom RPCs but I think this is good!

PR #658, opened 2023-05-10
update moonbeam randomness deposit amount

eshaben:
### Description

This PR updates the randomness deposit amount for Moonbeam, which is 100 GLMR


albertov19 (approved):
LGTM

PR #663, opened 2023-05-12
Add v0.8.20 note

eshaben:
### Description

This PR updates all of the tutorials where users could run into problems with v0.8.20. I've tested out all of the configs and added in the specific errors as they are formatted in case a user searches for the error.

~~Still need to address the Ethers.rs page, but going to recruit some help as I tried but I just don't know enough about rust 🤷‍♀️~~ The Ethers.rs page is configured for version 1.0.2, which supports compilers up to London, so we don't need to worry about this for now. There is a ticket to update this page to version 2.x

### Checklist


builders/build/eth-api/dev-env/brownie.md
```diff
@@ -204,14 +204,17 @@ If you want to specify the compiler version or compilation settings, you can do
 
 ```yaml
 compiler:
-  evm_version: london
+  evm_version: null
```
albertov19, line 207: Why setting this to null?

albertov19 (approved):
LGTM!

builders/build/eth-api/dev-env/brownie.md
eshaben: cause this will default to the latest evm_version that brownie supports, which as of right now is istanbul. 

but i figured i would leave it there so it shows the user they can update the evm version there and then the !!! note below points to the supported versions, in case brownie adds support for some of the newer compiler versions at some point

PR #666, opened 2023-05-16
update docs based on feedback from kapa: transaction priority and native token ERC-20s

eshaben:
### Description

This PR updates the docs based on feedback from kapa:

- Transaction priority (verified with Telmo)
- Native token ERC-20s

### Checklist


albertov19 (approved):
Love it!!!

PR #671, opened 2023-05-26
Westend fixes for XCM

albertov19:
### Description

We fixed this in some parts but not all. This ensure that we use the `DescendOrigin` now with how XCM V3 does it.

@eshaben could you please add this PR to the corresponding translation ticket :)

### DIFF
--- a/builders/interoperability/xcm/remote-evm-calls.md
+++ b/builders/interoperability/xcm/remote-evm-calls.md
@@ -134,7 +134,7 @@ When the XCM instruction gets executed in Moonbeam (Moonbase Alpha in this examp
     interior: {
       X1: {
         AccountId32: {
-          network: {westend: null },,
+          network: {westend: null },
           id: decodedAddress,
         },
       },
--- a/builders/interoperability/xcm/remote-evm-calls.md
+++ b/builders/interoperability/xcm/remote-evm-calls.md
@@ -110,7 +110,7 @@ For example, from the relay chain, the [`DescendOrigin`](https://github.com/pari
   DescendOrigin: {
     X1: {
       AccountId32: {
-        westend: null,
+        network: {westend: null },
         id: decodedAddress,
       },
     },
@@ -134,7 +134,7 @@ When the XCM instruction gets executed in Moonbeam (Moonbase Alpha in this examp
     interior: {
       X1: {
         AccountId32: {
-          westend: null,
+          network: {westend: null },,
           id: decodedAddress,
         },
       },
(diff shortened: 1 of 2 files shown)


builders/interoperability/xcm/remote-evm-calls.md
```diff
@@ -110,7 +110,7 @@ For example, from the relay chain, the [`DescendOrigin`](https://github.com/pari
   DescendOrigin: {
     X1: {
       AccountId32: {
-        network: 'Westend',
+        westend: null,
```
eshaben, line 113: it works as it was, but with this change the network isn't filled in correctly. reference encoded call data: 0x630303040b0101000c36e9ba26fa63c60ec728fe75fe57b86a450d94e7fee7f9f9eddd0d3f400d670000

albertov19: That should not be right. This is what or how the origin chain builds the Descend Origin instruction.

This is the XCM message Moonbase receives via UMP/DMP:

```
yarn xcm-decode-para --w wss://wss.api.moonbase.moonbeam.network --b 4430157 --channel dmp

...
Descend Origin:
{"descendOrigin":{"x1":{"accountId32":{"network":{"westend":null},"id":"0x6e3a195f4a67e2e6e96cd5f2070f73f6e66068db3dcbee136c7f3356523c4c76"}}}}
```

So even if you build the instruction in the Relay chain, this is how it is received on the parachain

eshaben: ok then, you would have to change this to `{"network": "westend":null}` instead of `"westend:null"`?

albertov19: Yeah ok good point, I've addressed this

builders/interoperability/xcm/remote-evm-calls.md
```diff
@@ -110,7 +110,7 @@ For example, from the relay chain, the [`DescendOrigin`](https://github.com/pari
   DescendOrigin: {
     X1: {
       AccountId32: {
-        network: 'Westend',
+        network: {westend: null },
```
eshaben, line 113:
```suggestion
        network: { westend: null },
```

builders/interoperability/xcm/remote-evm-calls.md
```diff
@@ -134,7 +134,7 @@ When the XCM instruction gets executed in Moonbeam (Moonbase Alpha in this examp
     interior: {
       X1: {
         AccountId32: {
-          network: 'Westend',
+          network: {westend: null },
```
eshaben, line 137:
```suggestion
          network: { westend: null },
```

eshaben (approved):

PR #682, opened 2023-06-23
add v1 docs for the xcm sdk

eshaben:
### Description

This PR adds new docs for the XCM SDK v1. Like with v0, it is split up by reference and usage. It aims to have better coverage on the reference page by including parameters and returned data.

### Checklist

- [x] If pages have been moved around, I have created an additional PR in `moonbeam-mkdocs` to update redirects

- [x] Will need to create PR in `moonbeam-mkdocs` repo to add redirects for Chinese site


builders/interoperability/xcm/xcm-sdk/v1/reference.md
```diff
@@ -0,0 +1,315 @@
...
+    |       Name       |      Type      |                       Description                       |
+    |:----------------:|:--------------:|:-------------------------------------------------------:|
+    |     `asset`      |    *Asset*     |            The asset's key and origin symbol            |
+    |   `balanceId`    | *ChainAssetId* |  The balance ID of the asset. Defaults to the asset ID  |
+    |    `decimals`    |    *number*    |          The number of decimals the asset has           |
+    |       `id`       | *ChainAssetId* |                      The asset ID                       |
+    |   `metadataId`   | *ChainAssetId* |              The metadata ID of the asset               |
+    |     `minId`      | *ChainAssetId* |               The minimum ID of the asset               |
+    | `palletInstance` |    *number*    | The number of the pallet instance the asset belongs to  |
+    |      `min`       |    *number*    | The minimum amount of the asset that can be transferred |
```
eshaben, line 74: @ekenigs this section in particular might need some help, haha  do you have better descriptions for some of these items?

builders/interoperability/xcm/xcm-sdk/v1/xcm-sdk.md
```diff
@@ -0,0 +1,691 @@
...
+
+=== "Minimum"
+
+    ```js
+    ...
+
+    const amount = data.min.toDecimal();
+    const symbol = data.min.originSymbol;
+
+    amount == 0n
```
ekenigs, line 600: I think this condition is wrong.

builders/interoperability/xcm/xcm-sdk/v1/xcm-sdk.md
```diff
@@ -0,0 +1,691 @@
...
+
+=== "Maximum"
+
+    ```js
+    ...
+
+    const amount = data.max.toDecimal();
+    const symbol = data.max.originSymbol;
+
+    amount == 0n
```
ekenigs, line 613: Same.

builders/interoperability/xcm/xcm-sdk/v1/xcm-sdk.md
```diff
@@ -0,0 +1,691 @@
...
+```js
+...
+const sourceChain = data.source.chain.name;
+const sourceFee = data.source.fee;
+
+const destinationChain = data.destination.chain.name;
+const destinationFee = data.destination.fee;
+
+console.log(
+  `You will pay ${data.source.fee.toDecimal()} ${
```
ekenigs, line 657: You forgot to use variables here instead of accessing it on `data`.

ekenigs (changes requested):

builders/interoperability/xcm/xcm-sdk/v1/reference.md
```diff
@@ -0,0 +1,315 @@
...
+    | `assetsData`  | *Map<string, ChainAssetsData>* |                           A list of the assets that the chain supports                           |
+    | `genesisHash` |            *string*            |                                  The hash of the genesis block                                   |
+    | `parachainId` |            *number*            |                                     The ID of the parachain                                      |
+    | `ss58Format`  |            *number*            | The [ss58 format](https://polkadot.js.org/docs/keyring/start/ss58/){target=_blank} for the chain |
+    |     `ws`      |            *string*            |                               The WebSocket endpoint for the chain                               |
+    |     `id`      |            *number*            |                            **For EVM parachains only** - The chain ID                            |
+    |     `rpc`     |            *string*            |                **For EVM parachains only** - The HTTP RPC endpoint for the chain                 |
+
+- `ChainAssetsData` - defines information needed to target the asset on the chain
```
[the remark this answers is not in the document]
eshaben: Ok I've tried to make some sense of this and apply a generic statement here, but not sure if I did it effectively. Please review this in particular again and let me know if and how I can improve

builders/interoperability/xcm/xcm-sdk/v1/reference.md
```diff
@@ -0,0 +1,315 @@
...
+
+    ```ts
+    type ChainAssetId =
+      | string
+      | number
+      | bigint
+      | { [key: string]: ChainAssetId };
+    ```
+
+### Transfer Data {: #transfer-data }
```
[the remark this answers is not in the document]
eshaben: I've got some questions...

So, if the top level `min` amount is the minimum amount of the asset that can be transferred, and the `max` is the maximum amount of the asset that can be transferred then...

The source `max` differs from the top level `max` how?

The source `min` you said is the min amount to keep on the source chain? Due to existential deposit-like circumstance where an asset is not native? Does that apply here and that is why there is an existential deposit and min?

The destination min is the minimum amount that can be transferred? But how does that differ from the top level `min`?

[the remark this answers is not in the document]
ekenigs: > The source max differs from the top level max how?

Here you are right the `max` is the same. I copied it to the top level just so it's easier to access and users don't need to remember if it's in `source` or in `destination`.

> The destination min is the minimum amount that can be transferred? But how does that differ from the top level min?

Top lvl `min` is different than destination `min` and it's a bit more complex. Formula is:


```
// isTheSameAsset - is a condition where we check that asset what we send is the same with `min`, `existentialDeposit`, `fee`
min = (isTheSameAsset ? fee : 0) + (isTheSameAsset ? existentialDeposit : 0) + (isTheSameAsset ? min : 0)
```
The code itself looks like this.

```typescript
  const result = Big(0)
    .plus(balance.isSame(fee) ? fee.toBig() : Big(0))
    .plus(
      balance.isSame(existentialDeposit) &&
        balance.toBig().lt(existentialDeposit.toBig())
        ? existentialDeposit.toBig()
        : Big(0),
    )
    .plus(balance.toBig().lt(min.toBig()) ? min.toBig() : Big(0));
```

ekenigs (approved):
It looks good from my side. Good job as usual. 👍

albertov19 (approved):
LGTM!

PR #684, opened 2023-06-27
Add docs on the multisig pallet

eshaben:
### Description

This PR includes docs on the multisig pallet to the pallets section. It also includes some docs on how to use the multisig pallet. I know we're trying to step away from using Polkadot.js Apps, but I felt like this was an exception to that. Additionally, I think the tutorial section best fits on the pallet page, but let me know if you think it would be better under the learn section or elsewhere 🙂 

### Checklist


albertov19 (approved):
Great job here!!! All good from my side

PR #685, opened 2023-06-27
fix moonriver support curve percentages for canceller track

eshaben:
### Description

The support curve percentages for the canceller track were updated to align with the percentages discussed in the original proposal.

https://github.com/PureStake/moonbeam/pull/2281
https://forum.moonbeam.foundation/t/proposal-opengov-integration-parameter-discussion/259/19


albertov19 (approved):
LGTM!

PR #689, opened 2023-06-28
Add section on removing session keys

eshaben:
### Description

This PR adds a section to the collator account management page on how to remove session keys for collators who no longer want to be a collator

### Checklist


albertov19 (approved):
Maybe we should recommend collators to officially stop collating and leave the collator set before they remove the keys, if not, I'm not sure what would happen and if they are still occupying a collator slot?

PR #690, opened 2023-06-28
add note that the send example is for demo purposes

eshaben:
### Description

Add a note to the Send an XCM Message Cross-Chain section of the Send and Execute XCM Messages page that informs users that the example won't work and is for demo purposes only.

### Checklist


albertov19 (approved):
LGTM

PR #714, opened 2023-08-09
remove variables from snippets

eshaben:
### Description

remove variables from snippets. don't need to worry about updating the chinese hardhat tutorials quite yet, but will add a note in the translation tickets referencing this PR so it doesn't slip through the cracks.


themacexpert (approved):
LGTM!!

PR #715, opened 2023-08-09
fix checking prereqs section of batch tutorial

eshaben:
### Description

Not sure why but the spacing caused the second bullet item to appear as a header. This PR fixes that


themacexpert (approved):
LGTM

PR #721, opened 2023-08-18
Update note flags

eshaben:
### Description

This PR updates the node flags to remove the following unnecessary flags, as they are the default:

--execution wasm \
--wasm-execution compiled \

It also updates the state-trie-cache to 1GB as the recommended starting point and adds the sync flag to the list of common flags.

### Checklist


albertov19 (approved):
LGTM

PR #731, opened 2023-08-30
update links and remove mrl banner image

eshaben:
### Description

- The banner used for the MRL page was deleted in a separate PR and so it resulted in a 404. Just deleted the reference to the banner since it's being removed with the docs redesign anyway
- Fix 404 link
- Fix meta refresh redirects

### Corresponding PRs

https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/329


albertov19 (approved):
LGTM!

PR #732, opened 2023-08-30
Update Polkadot Repos Polkadot-SDK name

themacexpert:
### Description

Polkadot recently migrated / consolidated their repos under a single Polkadot-SDK name, so it was necessary to rename repos otherwise they would be pointing at the archives. Some links were not migrated, such as issue links - those were not migrated to the new repos so the links point to the old archive repos still. 

Description of the change: https://forum.polkadot.network/t/psa-parity-is-currently-working-on-merging-the-polkadot-stack-repositories-into-one-single-repository/2883

### Checklist

### Corresponding PRs

#### Items to be Updated


eshaben (commented):
ah this change is a nightmare for the xcm fees page. is there a way to track the version like we used to? cause sometimes changes are done to `master` but haven't been released so if anyone clicks on the links and the numbers don't match up that's not good. Or if we forget to update everytime a new runtime is released, and the numbers change they won't match up. We need to think about possible solutions for this

eshaben (approved):
LGTM! Thanks for the effort here 💜

PR #737, opened 2023-09-01
move macos snippets to be expandable & update tracing tag

eshaben:
### Description

The macOS snippets have needed to be manually updated for every client version (couldn't use variables in snippets), so I've decided to move them to an expandable snippet format so we can just rely on changing the variable. 

This also updates to the latest client version, v0.33.0

### Corresponding PR

https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/332


themacexpert (approved):
LGTM 👍

PR #750, opened 2023-09-19
switch development nodes back to v0.32.2

eshaben:
### Description

Use client v0.32.2 for development nodes since there is a bug with v0.33.0 that impacts a development nodes ability to produce blocks. This bug will be fixed in the next client release.


themacexpert (approved):
LGTM 👍

PR #768, opened 2023-10-06
Minor Hardhat Updates

themacexpert:
### Description

- `npx hardhat` is deprecated and should be replaced with `npx hardhat init` 
- I also noticed a missing license identifier so I added that. 

### Corresponding PRs

CN: https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/342


eshaben (approved):
LGTM

PR #779, opened 2023-10-30
remove gelato docs and relayers section

eshaben:
### Description

Gelato no longer supports Moonbeam, so this PR removes the Gelato docs and relayers section
Goes with CN PR: https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/354


themacexpert (approved):
LGTM 🚢

PR #780, opened 2023-10-30
update the web3.js docs

eshaben:
### Description

This PR updates the Web3.js examples for v4.x, basically the only change we need to implement is defining the `gasPrice` and `nonce`. Pretty awesome for a major update!

Goes with CN PR: https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/355

### DIFF
--- a/.snippets/code/builders/build/eth-api/libraries/web3-js/balances.js
+++ b/.snippets/code/builders/build/eth-api/libraries/web3-js/balances.js
@@ -1,4 +1,4 @@
-const Web3 = require('web3');
+const { Web3 } = require('web3');
 
 // 1. Add the Web3 provider logic here:
 const providerRPC = {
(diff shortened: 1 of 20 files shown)


themacexpert:
There was 1 more change needed so I pushed a quick commit, I think I got everything but good to double check perhaps :)

https://docs.web3js.org/guides/web3_upgrade_guide/x/#the-web3-package-will-return-all-web3js-exported-objects

No changes are required for import web3 only when we do const { Web3 } = require('web3');

themacexpert (approved):
LGTM 🚢

eshaben:
> There was 1 more change needed so I pushed a quick commit, I think I got everything but good to double check perhaps :)
> 
> https://docs.web3js.org/guides/web3_upgrade_guide/x/#the-web3-package-will-return-all-web3js-exported-objects
> 
> No changes are required for import web3 only when we do const { Web3 } = require('web3');

Thanks! Figured I would change the imports anyways 🙂

PR #786, opened 2023-11-10
Update coinbase wallet docs and connect wallet button

eshaben:
### Description

This PR updates the Coinbase Wallet docs. I tested out the Coinbase Wallet mobile app and the browser extension; both flows are comparable, so I opted out of creating two sections because the one on the mobile app can easily be adapted for the browser extension.

I also updated the connect wallet button and tested out every scenario i could think of (i.e., user doesn't have moonbeam configured at all, user does have moonbeam configured but doesn't have an account connected, etc.).

This PR goes with the following mkdocs PR that updates button text and fixes a few errors: https://github.com/papermoonio/moonbeam-mkdocs/pull/153

### Checklist

#### Items to be Updated

All of the new images are listed under the `new` directory, so after translations, we'll need to remove the old and replace with the new!


albertov19 (approved):
LGTM!

You can actually send and receive DEV tokens. I tested it out and it works.

eshaben:
> LGTM!
> 
> You can actually send and receive DEV tokens. I tested it out and it works.

On the mobile app you can send DEV tokens? How? haha I didn't see a way that worked

PR #805, opened 2023-12-04
update identity precompile (on moonbase only)

eshaben:
### Description

This PR updates the identity precompile docs as its only available on moonbase alpha right now

### Checklist

- [x] Will need to create PR in `moonbeam-mkdocs` repo to add redirects for Chinese site


albertov19 (approved):
LGTM

PR #807, opened 2023-12-06
deprecate the remix plugin

eshaben:
### Description

This PR removes the Remix Plugin section from the Remix docs since we are deprecating the Remix plugin
Goes with CN PR: https://github.com/moonbeam-foundation/moonbeam-docs-cn/pull/369


themacexpert (approved):
LGTM 👍

PR #841, opened 2024-01-09
add a section on marking collators as inactive

eshaben:
### Description

This PR adds a section on marking a collator as inactive to the collator docs.

### Checklist


node-operators/networks/collators/activities.md
```diff
@@ -146,54 +147,54 @@ If you want to temporarily leave the candidate pool, you can easily do so using
...
+As a candidate, there are two options for increasing one's stake. The first and recommended option is to send the funds to be staked to another owned address and [delegate them to your collator](/tokens/staking/stake/#how-to-nominate-a-collator). Alternatively, collators that already have at least the [minimum self-bond amount](/node-operators/networks/collators/requirements/#minimum-collator-bond){target=_blank} staked can increase their bond through [Polkadot.js Apps](https://polkadot.js.org/apps/?rpc=wss://wss.api.moonbase.moonbeam.network#/accounts). Navigate to the **Developer** tab, click on **Extrinsics**, and follow these steps:
```
themacexpert, line 169: We should include a warning here that increasing your collator bond, while possible, does not increase rewards for collators. The only way to increase their rewards is to stake funds from another address. I'm not aware of any situations in which a collator would want to increase their bond without getting additional rewards.

themacexpert (approved):
LGTM 👍

node-operators/networks/collators/activities.md
eshaben: Wouldn't one reason be if the collator was trying to make it into the active set and needed to increase their total bond amount?

albertov19: @themacexpert @eshaben I believe that collator's bond account as self-stake. Meaning that if they increase their bond, it is accounted as them increasing the stake in theirself

https://github.com/moonbeam-foundation/moonbeam/blob/master/pallets/parachain-staking/src/lib.rs#L1937-L1952

eshaben: Yeah, that's what I thought. That it would increase their self-bond, therefore their total bond. I wasn't 100% sure what impact increasing their self-bond amount had on their rewards

PR #856, opened 2024-01-25
remove termynal js

eshaben:
### Description

We're not using any of the termynal animations, and, for the sake of performance, we don't need to either. So this PR removes the extra unnecessary JavaScript file.

### Checklist


albertov19 (approved):
LGTM

PR #862, opened 2024-02-02
remove the init kapa file

eshaben:
### Description

This PR removes the init kapa js file, as it is now being rendered inline in the HTML files
Goes with mkdocs PR: https://github.com/papermoonio/moonbeam-mkdocs/pull/167

### Checklist


albertov19 (approved):
I'll approve but we need to ensure the problems with PR https://github.com/papermoonio/moonbeam-mkdocs/pull/167 is fixed first before this is merged 😄

PR #863, opened 2024-02-05
resize moonbot image

eshaben:
### Description

This PR just resizes the moonbot image for the kapa widget so that we're serving properly sized images and reducing the overall bytes to be consumed by the network.

Goes with https://github.com/papermoonio/moonbeam-mkdocs/pull/168

### Checklist


albertov19 (approved):
94% down? wow xD

PR #864, opened 2024-02-07
Add ape docs

eshaben:
### Description

This PR adds docs for Ape. Since Brownie is no longer being maintained, I figured we should have docs for an alternative solution for python developers.

### Checklist

- [x] I have run my changes through Grammarly

### DIFF
--- a/builders/build/eth-api/dev-env/ape.md
+++ b/builders/build/eth-api/dev-env/ape.md
@@ -1,0 +1,333 @@
+---
+title: Deploy Contracts with Ape
+description: Use Ape, a Python framework, to compile, deploy, and debug smart contracts using Python on Moonbeam, thanks to its Ethereum compatibility.
+---
+
+# Using Ape to Deploy To Moonbeam
+
+## Introduction {: #introduction }
+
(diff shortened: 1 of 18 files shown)


albertov19 (dismissed):
This looks great as is!

One minor comment would be on doing something around testing??

Wdyt?

builders/build/eth-api/dev-env/ape.md
```diff
@@ -121,6 +121,52 @@ ape compile
...
+
+```bash
+touch tests/conftest.py
+```
+
+Since your tests will rely on the injection of the fixtures, you must define the fixtures first. When defining fixtures, you need to apply the `pytest.fixture` decorator above each function. For this example, you'll create two fixtures: one that defines the owner of the contract and one that deploys the contract from the owner's account.
+
+The `owner` fixture will use the built-in `accounts` fixture to take the first account in the list of test accounts provided by Ape and return it. The `box` fixture will deploy the `Box` contract type using the built-in `project` fixture, you simply have to provide the name of the contract and use the `owner` fixture to deploy it.
+
+```python title="tests/conftest.py"
```
albertov19, line 146: This is also really cool!

PR #876, opened 2024-02-27
Bump to client v0.36.0

eshaben:
### Description

This PR just updates the client version to v0.36.0!

### Checklist


albertov19 (changes requested):
Note that runtime might be 2801

eshaben:
> Note that runtime might be 2801

This is just for the client version. If we bump to runtime 2801 and there are new docker images at that point, I'll create a new PR. Aside from that, there will be separate PRs for each network when the network upgrades to 280x

albertov19:
Ok cool I was just thinking that some logs contain runtime 2800 word. But no biggie

albertov19 (approved):
LGTM

eshaben:
> Ok cool I was just thinking that some logs contain runtime 2800 word. But no biggie

ah ok, yeah I'll make sure to update those logs once 2801 is released

PR #918, opened 2024-05-02
April Ahrefs: Update links

eshaben:
### Description

We were getting some warnings about redirect pages, and I realized that it's because if you go to our site using this url https://docs.moonbeam.network/builders it will redirect and append a backslash so that the url is https://docs.moonbeam.network/builders/

To avoid a need for that, I've added backslashes to all links, as it was actually less time consuming then going one by one haha 😅 

This also fixes internal links pointing to redirects.

### Checklist


albertov19 (approved):
LGTM

albertov19 (approved):
LGTM

PR #922, opened 2024-05-10
link cleanup

eshaben:
### Description

I ended up updating and merging a bulk change to the links, but wasn't very smart about it. I didn't account for a lot of the little nuances and then I tried to backtrack to account for them, and I feel like I didn't catch all of the updates. So this PR is just to wipe the slate clean and start over with a more accurate approach that I feel way more confident with.

👉 The first commit finds all links that end in a backslash and removes the backslash
👉 The second commit runs a script that updates all internal links that do not reference images or sections and adds a backslash to those links
👉 The third and fourth commit cleans up internal links so they don't include https://docs.moonbeam.network (as they shouldn't for internal links) and updates all references to the docs site URL that are not actually links (i.e., in disclaimers, etc.)

Associated mkdocs PR that includes the script: https://github.com/papermoonio/moonbeam-mkdocs/pull/187

### Checklist


albertov19 (approved):
LGTM

themacexpert (approved):

PR #1086, opened 2025-01-24
Resolve issue with the MetaMask button

0xlukem:
### Description

I added a missing event listener for the `.connectMetaMask` button in the body so it calls the same `connectNetwork `function used by the nav button. This ensures that the body button now correctly connects to the specified network.

### Checklist

- [ ] I have added a label to this PR 🏷️
- [ ] I have run my changes through Grammarly

### DIFF
--- a/js/connectMetaMask.js
+++ b/js/connectMetaMask.js
@@ -113,7 +113,7 @@ connectMetaMaskNav.addEventListener('click', async (e) => {
       await displayConnectedAccount(connectedMoonbeamNetwork, connectedMoonbeamNetworkButton);
     }
   } else {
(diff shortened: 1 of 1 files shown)


themacexpert (approved):
I had to clear cookies but after that it worked fine! LGTM

eshaben (commented):
Sweeeet! It's working now!

On the Connect MetaMask page, after you connect to a network, the button text doesn't change to show "Connected <shortened_account_here>" and disables the button like it does in the modal. So ideally, that should be added in because if I am already connected to a network, but I try to click the button again, nothing happens. I don't think anyone would do this haha, but you never know. Anyways, so if the button was disabled, this issue would be avoided.

eshaben (approved):
LGTM! Thanks!

PR #1101, opened 2025-02-13
Remove DappRadar

themacexpert:
### Description

Dapp Radar is being decommed for moonbeam and moonriver networks next week, we should remove it from the doc site. 

### Checklist

### Corresponding PRs

#### Items to be Updated


eshaben (commented):
Do we want to add other dapps lists in the future? Just trying to determine if we should keep this as a section or not? Not the best user experience to click on dapps list, which leads to the index page, and then have to click on moonbeam dapps list to actually get to the dapps list

themacexpert:
Ready for another look

eshaben (commented):
Looks good! Just gotta figure out what's going on with the llms flow. Maybe try generating it again?

Oh! I almost forgot can you please open a PR to update the redirects.json file for dappradar and also the dapps list index page? And I actually think we might need to move some images just a little bit

themacexpert:
Replaced by https://github.com/moonbeam-foundation/moonbeam-docs/pull/1105

PR #1104, opened 2025-02-19
Correct list of pages populated in the llms.txt file

eshaben:
### Description

Correct the links by adding a slash at the end of the base URL

### Checklist

### DIFF
--- a/.github/workflows/check-llms.yml
+++ b/.github/workflows/check-llms.yml
@@ -21,6 +21,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install pyyaml requests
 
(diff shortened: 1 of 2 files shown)


eshaben:
Ah I actually just noticed that there is still a problem. We're listing the index pages as a doc page, which is fine, but they include `index` in the URL

albertov19:
> Ah I actually just noticed that there is still a problem. We're listing the index pages as a doc page, which is fine, but they include `index` in the URL

Pushed a commit to fix the index pages issue. Their content is included but for the URL the `index` at the end is removed. 

We need to push this to the other scripts

themacexpert (approved):
LGTM

PR #1116, opened 2025-03-20
3501 Runtime Upgrade Moonbeam

themacexpert:
### Description

3501 Runtime Upgrade

### Checklist


eshaben (commented):
What about the `spec_version` in the `variables.yml`? And checking any links that are using this to make sure they're still valid and pointing to the correct lines?

themacexpert:
> What about the `spec_version` in the `variables.yml`? And checking any links that are using this to make sure they're still valid and pointing to the correct lines?

added
All the spec version links look good

eshaben (approved):
LGTM!

PR #1200, opened 2025-08-04
Fix 404

0xlukem:
### Description

Quick fix on a URL 

### Checklist

- [ ] I have added a label to this PR 🏷️


eshaben (commented):
Was there a ticket created for the devrels to go back and update the other links then? And rollback the SDK version variable to the one that Moonbeam is currently using?

eshaben (approved):
For posterity, there was a ticket created to address the root of the issue in a separate PR. The goal of this one is just to resolve the broken link for now.

LGTM

